PR #3, opened 2018-06-08
Add GDS admin styling

irisfaraway:
https://eaflood.atlassian.net/browse/WC-126

We should install and use the standard styling created for admin systems by GDS.


spec/spec_helper.rb
```diff
@@ -54,44 +54,44 @@
   # is tagged with `:focus`, all examples get run. RSpec also provides
   # aliases for `it`, `describe`, and `context` that include `:focus`
   # metadata: `fit`, `fdescribe` and `fcontext`, respectively.
-  config.filter_run_when_matching :focus
+  # config.filter_run_when_matching :focus
```
Cruikshanks: Guess we need to have a chat about this, and also look to agree it across projects. I'm interested to know why the rspec default config has all been commented out? There maybe some that are unneeded but I believe there are a number that are valuable and important to keep.

irisfaraway:
@Cruikshanks Rspec changes reverted in anticipation of the next headbanging meeting 👍

PR #22, opened 2018-07-03
Mount renewals journey in back office

irisfaraway:
https://eaflood.atlassian.net/browse/WC-123

Users should be able to renew via an assisted digital route. The back office should support this.

Now that we have the WasteCarriersEngine, we can mount it in the back office to get the renewals journey.

Once we have the journey working in the app, we can start making any additional changes necessary for the AD process.


irisfaraway:
Ripping out all the user stuff (now provided by the engine) has reduced the total amount of code, which is what is setting CodeClimate off. I vote we merge it anyway.

Cruikshanks:
Oh, and meant to say agree on the codeclimate stuff!

PR #93, opened 2018-08-24
Users can approve or reject based on convictions

irisfaraway:
https://eaflood.atlassian.net/browse/WC-376

After reviewing convictions, a back office user should be able to approve or reject the renewal, and leave a comment explaining the decision.

Approving won't cause the renewal yet (see WC-302) but should update the conviction_sign_off.

Rejecting should change the status of the renewal, but leave the original registration alone.


irisfaraway:
Think https://github.com/DEFRA/waste-carriers-renewals/pull/252 needs to be merged in before this.

PR #95, opened 2018-08-30
Let user continue to renewal after logging in

irisfaraway:
https://eaflood.atlassian.net/browse/WC-448

If a user who isn't logged into the back office attempts to access a renewal, eg at /bo/renew/CBDU1234, they will be asked to log in. However, once they log in, they get redirected to the dashboard instead of the renewal.

This commit fixes this issue by adding the more complex path behaviour from the waste-carriers-front-office, so after logging in, users should be redirected to the resource they originally requested.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -15,8 +15,14 @@ class ApplicationController < ActionController::Base
     redirect_to bo_path
   end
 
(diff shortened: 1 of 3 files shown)


irisfaraway:
These don't fail locally 🤔 Investigating...

PR #96, opened 2018-08-30
Set up roles for back office users

irisfaraway:

There are a number of different roles in the existing system:

- agency user
- agency user with refund
- finance user
- finance admin
- agency super
- finance super

We need to replicate these in the new back office so we can assign permissions to them.


app/models/ability.rb
```diff
@@ -3,8 +3,67 @@
...
+    permissions_for_finance_admin_user if finance_admin_user?(user)
+    permissions_for_agency_super_user if agency_super_user?(user)
+    permissions_for_finance_super_user if finance_super_user?(user)
+  end
+
+  private
+
+  # Permissions for specific roles
+
+  def permissions_for_agency_user_group
```
Cruikshanks: Any particular reason for having a slightly different naming convention for these users e.g. `agency_user_group` rather than `agency_user`, which would fit more with the names used for other user types.

This will probably also answer my question on `in_agency_user_group?` rather than `agency_user?`

If its because we are trying to stick to the current names, then we can have a conversation because I think they are open for change 😁

irisfaraway: I agree it's a bit wonky. Basically the reason is because `agency_user_with_refund` can do all the things `agency_user` can, and also do refunds. However, since we haven't built any refunds in yet, they're currently the same.

However, when we add in users, we want to distinguish between the ones who should be able to have that refund power and those who should not.

Cruikshanks: I think I get it. So this saves us from having a `permissions_for_agency_refund_user` method which is exactly the same as `permissions_for_agency_user` apart from this one function. In this way we can apply the same permissions for all using `permissions_for_agency_user_group` and then just attach the refund permission to those who need the ability.

irisfaraway: @Cruikshanks Yep, that's the idea :)

PR #97, opened 2018-09-03
Finance users can record electronic bank transfer payment

irisfaraway:

A user with the finance or finance_admin role should be able to record when a bank transfer payment has been received. Recording the receipt of the payment should also update the balance.

This PR should lay the groundwork for recording payments of other types.


irisfaraway:
This is just about done – just awaiting the merger of https://github.com/DEFRA/waste-carriers-renewals/pull/256 so the last test passes.

PR #101, opened 2018-09-05
Back office users can select which type of payment they want to add

irisfaraway:

There are 5 different types of payment which can be manually added. The type of payment a user can add also depends on their role.

Rather than sending the user straight into a form, we first ask them to select which type of payment they want to record, and then direct them to the correct form for that type.


irisfaraway:
Once https://github.com/DEFRA/waste-carriers-back-office/pull/97 is merged in, we should be able to start testing successful submission of the form.

PR #115, opened 2018-09-14
Configure back-office to send emails via SendGrid

irisfaraway:
https://eaflood.atlassian.net/browse/WC-393
https://eaflood.atlassian.net/browse/WC-353
https://eaflood.atlassian.net/browse/WC-354

The engine will be responsible for building emails and triggering when they are sent (see DEFRA/waste-carriers-renewals#261 for implementation of the completed renewal confirmation email). However, as the host applications are sending the emails, they need to be configured to enable this.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -87,13 +87,12 @@ Rails.application.configure do
   config.action_mailer.raise_delivery_errors = false
   config.action_mailer.delivery_method = :smtp
 
-  # Default settings are for mailcatcher
   config.action_mailer.smtp_settings = {
     user_name: ENV["WCRS_EMAIL_USERNAME"],
     password: ENV["WCRS_EMAIL_PASSWORD"],
     domain: config.wcrs_back_office_url,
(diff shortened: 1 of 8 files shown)


config/environments/production.rb
```diff
@@ -83,4 +79,22 @@
...
   config.log_formatter = ::Logger::Formatter.new
+
+  # Sending e-mails is required for user management and registration e-mails
+  config.action_mailer.default_url_options = { host: config.wcrs_back_office_url, protocol: "http" }
+
+  # Don't care if the mailer can't send (if set to false)
+  config.action_mailer.raise_delivery_errors = false
+  config.action_mailer.delivery_method = :smtp
+
+  # Default settings are for mailcatcher
```
Cruikshanks: In production can we remove the default settings? Should we forget to set the env vars the error is likely to point more to that, rather than because it failed to send using the credentials provided.

irisfaraway: Fixed!

PR #130, opened 2018-09-25
Better renewal end screens for back office users

irisfaraway:
https://eaflood.atlassian.net/browse/WC-478

Currently the renewal_complete and renewal_received forms show the same content to all users regardless of which app they are mounted in.

Most of the time, this is fine - however, it does cause some problems on these two pages at the end of the renewals journey, as back office users require different links and information.

This relies on https://github.com/DEFRA/waste-carriers-renewals/pull/270


irisfaraway:
Engine updates merged in, so should work now. 👍

PR #141, opened 2018-10-10
Use shared displayable_address method for displaying addresses

irisfaraway:
The `address_lines` method is another example of duplication which can now be replaced by the method added in https://github.com/DEFRA/waste-carriers-renewals/pull/281


irisfaraway:
Note that this won't work until https://github.com/DEFRA/waste-carriers-renewals/pull/281 is merged in and the dependency is updated.

PR #143, opened 2018-10-11
Add rake task to fix addresses in incorrect format

irisfaraway:
https://eaflood.atlassian.net/browse/WC-500

Once we roll out our fix to make waste-carriers-renewals and waste-carriers-frontend save addresses from OS Places in the same format, we will still have some addresses which have already been saved.

Running this rake task should find the addresses and correct them, so all addresses are in a consistent format.

### DIFF
--- a/lib/tasks/addresses.rake
+++ b/lib/tasks/addresses.rake
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "rest-client"
+
 namespace :address do
   desc "Fix addresses set in the renewals process to match old data format"
   task update_from_os_places: :environment do
@@ -37,11 +39,7 @@ def update_address(old_address)
 end
 
 def build_updated_address(old_address)
-  # Send request to OS Places
-  address_results = WasteCarriersEngine::AddressFinderService.new(old_address.postcode).search_by_postcode
-
-  # Find the correct match - use UPRN :)
-  matching_address = address_results.find { |result| result["uprn"] == old_address.uprn.to_s }
+  matching_address = search_by_uprn(old_address.uprn)
 
   if matching_address.blank?
     parent = old_address._parent
@@ -55,6 +53,36 @@ def build_updated_address(old_address)
   new_address.address_type = old_address.address_type
 
   new_address
+end
+
+def search_by_uprn(uprn)
+  url = "#{Rails.configuration.os_places_service_url}/addresses/#{uprn}"
+
+  begin
+    response = RestClient::Request.execute(
+      method: :get,
+      url: url
+    )
+
+    begin
+      JSON.parse(response)
+    rescue JSON::ParserError => e
+      Airbrake.notify(e) if defined?(Airbrake)
+      Rails.logger.error "OS Places JSON error: " + e.to_s
+      nil
+    end
+  rescue RestClient::BadRequest
+    Rails.logger.debug "OS Places: resource not found"
+    nil
+  rescue RestClient::ExceptionWithResponse => e
+    Airbrake.notify(e) if defined?(Airbrake)
+    Rails.logger.error "OS Places response error: " + e.to_s
+    nil
+  rescue SocketError => e
+    Airbrake.notify(e) if defined?(Airbrake)
+    Rails.logger.error "OS Places socket error: " + e.to_s
+    nil
+  end
 end
 
 def replace_old_address(old_address, new_address)
--- a/lib/tasks/addresses.rake
+++ b/lib/tasks/addresses.rake
@@ -22,8 +22,8 @@ end
 
 def update_addresses_for(registrations)
   registrations.each do |registration|
-    update_address (registration.registered_address) if address_is_from_os_places?(registration.registered_address)
-    update_address (registration.contact_address) if address_is_from_os_places?(registration.contact_address)
+    update_address(registration.registered_address) if address_is_from_os_places?(registration.registered_address)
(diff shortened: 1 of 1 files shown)


irisfaraway:
@Cruikshanks Think this would benefit from testing against more accurate data before use, but seems to work for ones I've generated locally. Have a look and see what you think!

lib/tasks/addresses.rake
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+namespace :address do
+  desc "Fix addresses set in the renewals process to match old data format"
+  task update_from_os_places: :environment do
+    registrations = WasteCarriersEngine::Registration.where(:past_registration.exists => true)
```
Cruikshanks: I would suggest instantiating these within the methods themselves. What we have makes the methods themselves more flexible and easier to test, but there may be a performance hit.

We have to hold both collections in memory until both operations have completed. Moving the instantiating to the relevant methods moves the variable to a more local scope, and will allow the GC to get in and scrub it from memory if it needs to.

lib/tasks/addresses.rake
```diff
@@ -0,0 +1,68 @@
...
+end
+
+def update_addresses_for(registrations)
+  registrations.each do |registration|
+    update_address_if_necessary(registration.registered_address)
+    update_address_if_necessary(registration.contact_address)
+  end
+end
+
+def update_address_if_necessary(address)
```
Cruikshanks: Is there a particular need for this method? Could we not just make it this in `update_addresses_for()`

```ruby
  update_address(address) if address_is_from_os_places?(registration.registered_address)
  update_address(address) if address_is_from_os_places?(registration.contact_address)
```

lib/tasks/addresses.rake
```diff
@@ -0,0 +1,68 @@
...
+def update_address(old_address)
+  new_address = build_updated_address(old_address)
+  return unless new_address
+
+  replace_old_address(old_address, new_address)
+end
+
+def build_updated_address(old_address)
+  # Send request to OS Places
+  address_results = WasteCarriersEngine::AddressFinderService.new(old_address.postcode).search_by_postcode
```
Cruikshanks: As this is just a temporary rake task we can get away with being a bit more messy with the code. We could just create a method here like the following

```ruby
def search_by_uprn(uprn)
      url = "#{Rails.configuration.os_places_service_url}/addresses/#{uprn}"

      begin
        response = RestClient::Request.execute(
          method: :get,
          url: url
        )

        begin
          JSON.parse(response)
        rescue JSON::ParserError => e
          Airbrake.notify(e) if defined?(Airbrake)
          Rails.logger.error "OS Places JSON error: " + e.to_s
          :error
        end
      rescue RestClient::BadRequest
        Rails.logger.debug "OS Places: resource not found"
        :not_found
      rescue RestClient::ExceptionWithResponse => e
        Airbrake.notify(e) if defined?(Airbrake)
        Rails.logger.error "OS Places response error: " + e.to_s
        :error
      rescue SocketError => e
        Airbrake.notify(e) if defined?(Airbrake)
        Rails.logger.error "OS Places socket error: " + e.to_s
        :error
      end
 end
```
We only have to wait for the one result to come back, and we don't have to search through any results.

irisfaraway: That was my initial thought but I was struggling to get the single-address requests to work locally as it appeared to be timing out. I'll have another bash though and see if I can get it to work.

Cruikshanks: Hopefully it should work. I mimicked doing this in IRB within the Vagrant box using the following if that helps

```
irb(main):001:0> require 'rest-client'
=> true
irb(main):002:0> url = "http://localhost:8005/addresses/340116"
=> "http://localhost:8005/addresses/340116"
irb(main):003:0> response = RestClient::Request.execute(
irb(main):004:1* method: :get,
irb(main):005:1* url: url
irb(main):006:1> )
=> <RestClient::Response 200 "{\"moniker\":...">
irb(main):007:0> result = JSON.parse(response)
NameError: uninitialized constant JSON
	from (irb):7
	from /home/vagrant/.rbenv/versions/2.4.2/bin/irb:11:in `<main>'
irb(main):008:0> require 'json'
=> true
irb(main):009:0> result = JSON.parse(response)
=> {"moniker"=>"340116", "uprn"=>"340116", "lines"=>["NATURAL ENGLAND", "DEANERY ROAD"], "town"=>"BRISTOL", "postcode"=>"BS1 5AH", "easting"=>"358205", "northing"=>"172708", "country"=>"", "dependentLocality"=>"", "dependentThroughfare"=>"", "administrativeArea"=>"BRISTOL", "localAuthorityUpdateDate"=>"", "royalMailUpdateDate"=>"", "partial"=>"NATURAL ENGLAND, HORIZON HOUSE, DEANERY ROAD, BRISTOL, BS1 5AH", "subBuildingName"=>"", "buildingName"=>"HORIZON HOUSE", "thoroughfareName"=>"DEANERY ROAD", "organisationName"=>"NATURAL ENGLAND", "buildingNumber"=>"", "postOfficeBoxNumber"=>"", "departmentName"=>"", "doubleDependentLocality"=>""}
irb(main):010:0>
```

irisfaraway: I just ran into other issues with the address lookup when I was refreshing VCR cassettes so looks like it was a local issue! Restarted the VM and it works fine 👍

PR #153, opened 2018-10-22
Transfer a registration to an existing account

irisfaraway:
https://eaflood.atlassian.net/browse/WC-61

Sometimes NCCC will need to transfer ownership of a registration to a different account.

As a first step, we will set up transfer of a registration to an account which already exists.

### DIFF
--- a/app/services/registration_transfer_service.rb
+++ b/app/services/registration_transfer_service.rb
@@ -37,5 +37,8 @@ class RegistrationTransferService
 
   def send_confirmation_email
     RegistrationTransferMailer.transfer_to_existing_account_email(@registration).deliver_now
+  rescue StandardError => e
+    Airbrake.notify(e) if defined?(Airbrake)
+    Rails.logger.error "Registration transfer confirmation email error: " + e.to_s
   end
 end
--- a/app/services/registration_transfer_service.rb
+++ b/app/services/registration_transfer_service.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+class RegistrationTransferService
+  def initialize(registration)
+    @registration = registration
+    @transient_registration = find_transient_registration
+  end
+
+  private
+
+  def find_transient_registration
+    WasteCarriersEngine::TransientRegistration.where(reg_identifier: @registration.reg_identifier).first
+  end
+end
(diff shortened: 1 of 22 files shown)


irisfaraway:
@Cruikshanks Ready for a review! PR definitely beefier than I would have liked – if anything is unclear, DM me and I'll try to reply quickly.

Will need subsequent PRs to create new user accounts when transferring, and also ideally to refactor the form controllers.

app/services/registration_transfer_service.rb
```diff
@@ -0,0 +1,41 @@
...
+    ExternalUser.where(email: email).first
+  end
+
+  def update_account_email_for(registration)
+    registration.account_email = @recipient_user.email
+    registration.save!
+  end
+
+  def send_confirmation_email
+    RegistrationTransferMailer.transfer_to_existing_account_email(@registration).deliver_now
```
Cruikshanks: Should have said this in previous PR's, but can we wrap this in a `begin/rescue`. Sending emails where we're not offloading them to an ActiveJob can sometimes be flaky (based on experience with other services). At this point the transfer would have been successful but the NCCC user would not see that if sending the email failed. They would just see an error message.

Catching the exception and logging it in Errbit means *we* still know something went wrong, but the main thing is the user gets to see the confirmation that the transfer is complete so no confusion.

config/locales/registration_transfers.en.yml
```diff
@@ -0,0 +1,37 @@
...
+      registration_info:
+        heading: "About this registration"
+        labels:
+          account_email: "Current user account"
+          reg_identifier: "Registration number"
+          company_name: "Company name"
+      subheading_1: "Transfer the registration"
+      email_label: "Email address"
+      confirm_email_label: "Confirm email address"
+      button: "Transfer account"
```
Cruikshanks: As we are transferring just a single registration I think have the submit button labelled **Transfer account** would be confusing. I think simply labelling it **Transfer** is sufficient in the context of all the other information we are showing.

```suggestion
      button: "Transfer"
```

irisfaraway: 🤦‍♀️  You're totally right! That's definitely inaccurate.

app/services/registration_transfer_service.rb
irisfaraway: Good point, will sort this out (and make a note to fix it in our other mailers too).

irisfaraway: Issue for the other mailers in the engine: https://github.com/DEFRA/waste-carriers-renewals/issues/293

PR #161, opened 2018-11-01
Transfer a registration to a new account

irisfaraway:
https://eaflood.atlassian.net/browse/WC-61

Now that we can transfer a registration to an existing account, we should also be able to transfer it to a new one.

When the registration transfer form is submitted, if there is no account with the email address used, immediately create a new account and transfer the registration.

The new account should then receive an email notifying them of the change and inviting them to set their password.

---

WIP


app/services/registration_transfer_service.rb
```diff
@@ -27,18 +19,56 @@ def find_transient_registration
   def find_user(email)
```
Cruikshanks: I'd like to see this method name changed. `find_user` infers it'll search for a user with the matching email address and return it, but in fact it'll go on and complete the transfer, then send a status result back. 
```suggestion
  def transfer_existing_user(email)
```

app/services/registration_transfer_service.rb
```diff
@@ -27,18 +19,56 @@ def find_transient_registration
...
+    return unless ExternalUser.where(email: email).first
+
+    # If a matching user exists, transfer and return a success status
+    update_account_emails(email)
+    send_existing_account_confirmation_email
+
+    :success_existing_user
+  end
+
+  def create_new_user(email)
```
Cruikshanks: I'd like to see this method name changed. `create_new_user` infers it'll create a one using the email address provided and return it, but in fact it'll go on and complete the transfer, then send a status result back. 
```suggestion
  def transfer_new_user(email)
```

app/services/registration_transfer_service.rb
```diff
@@ -27,18 +19,56 @@ def find_transient_registration
...
+  end
+
+  def send_new_account_confirmation_email(token)
+    RegistrationTransferMailer.transfer_to_new_account_email(@registration, token).deliver_now
+  rescue StandardError => e
+    log_email_error(e)
+  end
+
+  def log_email_error(error)
+    Airbrake.notify(error) if defined?(Airbrake)
```
Cruikshanks: I'd like to see us start sending additional data such as the registration number when we send an airbrake notification
```suggestion
    Airbrake.notify(error, reg_identifier: @registration.reg_identifier) if defined?(Airbrake)
```
See https://github.com/airbrake/airbrake-ruby#airbrakenotify

app/services/registration_transfer_service.rb
```diff
@@ -27,18 +19,56 @@ def find_transient_registration
...
+    Rails.logger.error "Registration transfer new account confirmation email error: " + error.to_s
```
Cruikshanks: This method is used for errors with both existing and new users so the message needs to be more generic.
```suggestion
    Rails.logger.error "Registration transfer confirmation email error: " + error.to_s
```

PR #175, opened 2018-11-08
Add a quick version of anonymise email task

Cruikshanks:
The current version of the anonymise email task ensures

- all emails are anonymised
- that the link between user, registrations and transient registrations remains intact

However it taks forever! We have had the task running for more than a day and it is still yet to complete.

So if we really need this we have the option, however we also need a much faster version. This change adds an alternate version that simply replaces the domain of all email addresses with **example.com**. It runs the risk that it will lead to duplicate users e.g. john.smith@gmail.com and john.smith@yahoo.com both become john.smith@example.com.

However we currently do not have a better option to anonymise all emails in a reasonable amount of time.

### DIFF
--- a/lib/db/anonymise_email.rb
+++ b/lib/db/anonymise_email.rb
@@ -17,6 +17,10 @@ module Db
       @counts = Db.counts(@collections[:users])
       @counts[:id_increment] = 1
 
+      @counts[:users_updated] = 0
+      @counts[:registrations_updated] = 0
+      @counts[:transient_registrations_updated] = 0
+
       @paging = Db.paging(@counts[:total], 100)
     end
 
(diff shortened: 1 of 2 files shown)


Cruikshanks:
Having looked into this it seems there is actually no easy way to implement such an update in Mongodb. You cannot update a document based on taking an existing value from that document, without iteration. So for example you cannot do

```ruby
db.registrations.find.update_many("$set: { accountEmail: currentEmail })
```

Having looked into it [aggregation](https://docs.mongodb.com/manual/aggregation/) is the closest solution, but that involves creating a new collection, copying each document into it with the necessary changes, then dropping the old collection and renaming the new one. We could do all this but still only see a marginal improvement in speed.

Hence closing this PR as do not see it as an avenue to explore.

PR #176, opened 2018-11-09
Update anonymise email rake to use bulk write

Cruikshanks:
When running in our pre-production environment this task is taking more than a day to complete. We therefore started looking into performance improvements.

One idea was to take the risk of duplicates and simply replace the domain in all emails with **example.com**. However we then realised its not actually very easy to update a field based on an existing value in the document when using MongoDb. In fact up to version 3.4 it wasn't even possible.

So instead we found [bulk_write](https://docs.mongodb.com/ruby-driver/master/tutorials/ruby-driver-bulk-operations/). At the moment for each user we are calling 3 different queries;

- update all registrations with a matching account email
- update all transient_registrationsnm with a matching account email
- update the user's email

In each case we are sending a request, waiting for it to complete, and then MongoDb is sending back a result.

With bulk_write we create an array of update statements for each collection, and then apply them as one. So for every 100 users we are just firing one update per collection. The aim is that the reduction in network traffic will result in an improvement in performance.

### DIFF
--- a/lib/db/anonymise_email.rb
+++ b/lib/db/anonymise_email.rb
@@ -20,52 +20,59 @@ module Db
       @paging = Db.paging(@counts[:total], 100)
     end
 
-    def anonymise(debug = false)
+    def anonymise
       while @paging[:page_number] <= @paging[:num_of_pages]
-        anonymise_users
+        page_through_users
         @paging[:page_number] += 1
-        print_progress unless debug
+        print_progress
       end
     end
 
     private
 
-    def anonymise_users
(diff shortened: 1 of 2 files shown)


Cruikshanks:
We simply don't think attempting to do this against collections of the size found in production is     feasible from the app.

Hence we have instead re-written this task in Javascript (including the use of bulk write) which can be called from the mongo shell directly on the database server, which we hope will mean we can complete this task in a much shorter time.

Hence we are closing this PR.

PR #182, opened 2018-11-09
Add rake task to sync internal users

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-524

When the service went live on September 19 we ran a JS script to create users in the back office based on users that already existed in the old back end.

As we do not currently have the ability to maintain users in the back office (see WC-495) this still remains the only means to create users.

Having spoken with the NCCC team lead, just being able to re-sync the back office users with those in the backend and create any new ones would suffice.

Hence this change which adds a rake task to do just that.

### DIFF
--- a/lib/db/anonymise_email.rb
+++ b/lib/db/anonymise_email.rb
@@ -31,7 +31,7 @@ module Db
     private
 
     def anonymise_users
-      Db.paged_users(@paging).each do |user|
+      Db.paged_collection(@paging, @collections[:users]).each do |user|
         unless user["email"].end_with?("@example.com")
           result = anonymise_email(
             debug,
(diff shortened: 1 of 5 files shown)


Cruikshanks:
We had, but I think that work should go into a separate PR. This focuses on the syncing itself (with a rake task to allow you to actually kick it off). The work to build that into a view, display results etc I think needs to be done separately.

As to putting it into the lib folder, currently that decision was led by
- Initiated as a rake task so add shifting your code out of the task into a class in lib is a common pattern
- Not actually having a clear idea of what to put in `app/services`, and what to put in `lib`!

A quick read of this post <https://stackoverflow.com/questions/16159021/rails-service-objects-vs-lib-classes> suggests this should be a service. But again I want to push that off to a subsequent PR. I've tested this enough times over the weekend to be happy that it works, and we've got folks waiting for access to system so really need to ship.

When we start giving users access to this feature via the UI, I'm happy that's the point it needs to be strengthened with Unit tests.

Feel free to argue though!

PR #187, opened 2018-11-12
Allow super users to migrate users from backend

irisfaraway:
https://eaflood.atlassian.net/browse/WC-495

Super users can create users in the backend. However, as these are currently stored in different DB collections, a new backend user won't automatically have access to the backend.

We have created a rake task to sync backend users to the back office (https://github.com/DEFRA/waste-carriers-back-office/pull/182) but this will require an RFC every time it is run. So it's better to hand that power over to the back office super users, so they can easily add a new user and make sure they have access to both apps, without having to wait for the rake task to be run.


irisfaraway:
At this point the service has been set up and tested, so it's time to move on to the UI portion of the story.

Cruikshanks:
Think I've seen some of this code somewhere before 🤔 😁 !

PR #191, opened 2018-11-16
Improve dashboard search performance

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-531

In production our users have reported that when attempting to apply a filter in the dashboard, it becomes unresponsive and eventually leads to a "service unavailable" page.

Initial investigation has shown that the problem is each time a user searches the app is pulling back all transient registrations and then iterating through them,  checking for the matching criteria.

One watching the logs this actually results in the ORM Mongoid querying for each transient registration. Hence its not putting a massive load on MongoDB, but it is now taking so long that the browsers is timing out leading to the service unavailable page.

This change is about improving the performance of the search in the dashboard, such that it no longer times out, and ideally is a scalable solution i.e. we want be back here again in another 6 months.

### DIFF
--- a/app/services/transient_registration_search_service.rb
+++ b/app/services/transient_registration_search_service.rb
@@ -12,6 +12,17 @@ class TransientRegistrationSearchService
   def search(page)
     return WasteCarriersEngine::TransientRegistration.none if no_search_terms_or_filters?
 
+    # The criteria when add results in a AND search (not an OR). So for a
+    # renewal to be returned it must match all criteria selected. In some cases
+    # this would be impossible, for example `pending_payment` filters on
+    # renewals that have been submitted with a balance != 0. If you also
+    # selected `in_progress` then you'd get nothing because a renewal can't be
+    # both.
+    #
+    # Also note the way criteria works in mongoid is that until we actually
+    # attempt to access a renewal e.g. with `.first` or ``.each` criteria is
+    # just a hash of filters. So when we're merging here we are merging filters
+    # not selected renewals!
     criteria = WasteCarriersEngine::TransientRegistration.order_by("metaData.lastModified": :desc)
     criteria.merge!(WasteCarriersEngine::TransientRegistration.search_term(@term)) if @term.present?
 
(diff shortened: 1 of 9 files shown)


app/services/transient_registration_search_service.rb
```diff
@@ -0,0 +1,34 @@
...
+class TransientRegistrationSearchService
+  def initialize(term, in_progress, pending_payment, pending_conviction_check)
+    @term = term
+
+    @in_progress = in_progress
+    @pending_payment = pending_payment
+    @pending_conviction_check = pending_conviction_check
+  end
+
+  def search(page)
```
irisfaraway: This may be personal preference, but it might be useful to add a comment explaining that all active conditions must be met for an item be in `criteria` (eg AND, not OR). When I was trying to work this out, I discovered that `merge!` for hashes combines any items that appear in either hash being merged, but `merge!` for ActiveRecord (and I assume also Mongoid) only combines items which appear in _both_ lists. So it's not totally clear from the method name. Might save future people some googling. What do you think?

Cruikshanks: The first point about for an item to be returned in the results, it must match all criteria selected makes sense, along with a note that we allow you to select an invalid combination.

Not sure anything about the merge though is necessary. If I've understood how you say ActiveRecord works then this

```ruby
    criteria = WasteCarriersEngine::TransientRegistration.order_by("metaData.lastModified": :desc)
    criteria.merge!(WasteCarriersEngine::TransientRegistration.search_term(@term)) if @term.present?
```

would fail because the 2 criteria being merged are not the same??? As the code is written and infers that the second criteria is merged with the existing criteria, I'm not sure a comment as well is necessary.

But I might have misunderstood what the issue is.

irisfaraway: Ahh, I guess I thought of it merging the list of items rather than the list of criteria! Maybe my bad then. But I think if we're agreed about the comment then that's fine.

Cruikshanks: Got it! Now that makes sense. Have pushed a comment, let me know what you think.

irisfaraway: Awesome, makes way more sense now! (And I learned something about Mongoid.)

PR #197, opened 2018-11-22
Add convictions overview page

irisfaraway:
https://eaflood.atlassian.net/browse/WC-483

Back office users can currently check for convictions which require review by searching and filtering on the dashboard. However, the actual workflow is more complex, so we need a separate page with more features to support this.


irisfaraway:
@Cruikshanks Fair point! I was thinking I didn't want it to be too similar to the main dashboard (with search, etc) as it felt like replicating functionality and I didn't want people to get confused between the two pages.

But I appreciate that the filtering has been implemented inconsistently, which could be an issue. I think this is because when we first built the dashboard search it was possible to stack multiple filters (eg "I want renewals with a pending payment AND a pending conviction check") but all the options for the convictions are mutually exclusive. This will actually be the case for the dashboard as well so maybe the answer is to change how the filtering works on the dashboard.

PR #201, opened 2018-11-23
Fix err msg WCR Engine::VERSION already defined

Cruikshanks:
When running unit tests or rake taks in the back office the first thing that appears is a warning message.

```bash
/vagrant/waste-carriers-back-office/config/initializers/waste_carriers_engine.rb:1: warning: already initialized constant WasteCarriersEngine::VERSION
/vagrant/waste-carriers-renewals/lib/waste_carriers_engine/version.rb:2: warning: previous definition of VERSION was here
```

After looking at the project, and comparing it to others where we have used a shared engine, spotted that we defined an initializer for the Waste carriers engine. In the initializer we then specify the version number for the engine.

This is not something we do in other projects so removed the initializer and this solved the warning message.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DEFRA/waste-carriers-renewals
-  revision: 9de9bd7703d79746133c59da718e86aa82074a7d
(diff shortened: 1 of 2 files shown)


Cruikshanks:
I'll get the front office sorted as well.

PR #207, opened 2018-11-27
Change redirect paths after conviction check action

irisfaraway:
https://eaflood.atlassian.net/browse/WC-483

After consulting with NCCC about their workflow, they would prefer for the app to redirect to different pages after the a conviction check action occurs.

- "Accept" should redirect to the 'possible matches' list
- "Has relevant convictions" should redirect to the 'possible matches' list
- "Reject" should redirect to the 'checks in progress' list

### DIFF
--- a/app/controllers/conviction_approval_forms_controller.rb
+++ b/app/controllers/conviction_approval_forms_controller.rb
@@ -12,7 +12,8 @@ class ConvictionApprovalFormsController < AdminFormsController
     return unless super(ConvictionApprovalForm,
                         "conviction_approval_form",
                         params[:conviction_approval_form][:reg_identifier],
-                        { authorize_action: :authorize_action })
+                        { authorize_action: :authorize_action,
+                          success_path: convictions_path })
 
     update_conviction_sign_off
     renew_if_possible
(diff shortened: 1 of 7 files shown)


irisfaraway:
Hakiri warning is due to CVE-2018-16476, which was just published but is not related to this change - it'll be in the master branch as well.

app/controllers/conviction_approval_forms_controller.rb
```diff
@@ -27,4 +30,8 @@ def authorize_action(transient_registration)
   def update_conviction_sign_off
     @transient_registration.conviction_sign_offs.first.approve!(current_user)
   end
+
+  def success_path
```
Cruikshanks: I don't think this is needed. Now we pass the value in as a hash, we have the key `success_path:` to tell us what the value relates to.

app/controllers/conviction_rejection_forms_controller.rb
```diff
@@ -26,4 +29,8 @@ def authorize_action(transient_registration)
   def reject_renewal
     @transient_registration.conviction_sign_offs.first.reject!
   end
+
+  def success_path
```
Cruikshanks: I don't think this is needed. Now we pass the value in as a hash, we have the key `success_path:` to tell us what the value relates to.

irisfaraway:
@Cruikshanks Moved the options refactor into a separate branch as suggested (https://github.com/DEFRA/waste-carriers-back-office/pull/209). This branch now merges into that one just for an easier view of the diff, but will merge to master after #209 has.

Also fixed the `success_path` stuff.

PR #217, opened 2018-12-03
Add status in dashboard for stuck renewals

Cruikshanks:
https://eaflood.atlassian.net/browse/WC-533

WC-533 is linked to one of a number of examples we've had where renewals have gottent stuck. In all cases it was because the renewal was expired, however when we attempted to complete the renewal we hadn't spotted this leading to a failed completion.

In these cases the renewals have no pending payments or convictions, and have been submitted. Therefore the current dashboards view doesn't display a status.

To better highlight those that are stuck, and to indicate to the user that action needs to be taken to resolve it, this changes updates the view to ensure a status is shown for stuck renewals.

---

WIP


app/views/dashboards/index.html.erb
```diff
@@ -89,11 +89,14 @@
                       <strong class="app-c-tag app-c-tag--success">
                         <%= t(".results.statuses.pending_payment") %>
                       </strong>
-                    <% end %>
-                    <% if transient_registration.pending_manual_conviction_check? %>
+                    <% elsif transient_registration.pending_manual_conviction_check? %>
```
irisfaraway: 'Pending payment' and 'Pending conviction check' are not mutually exclusive conditions – sometimes a renewal can have both, and so should display both flags. However this would mean that only the payment flag would be shown.

Cruikshanks: 🤦‍♂️

Cruikshanks: Confirmed this is back to how it was, but including the stuck status by testing in pre-prod. Ping me if you want some registration numbers

PR #255, opened 2019-02-05
Align rspec test setup with WEX

Cruikshanks:
Whilst working on the WEX service where the code is based on WCR renewals, we have found issues with the initial Database cleaner setup, as well as how simplecov was configured.

We have also settled on a pattern where any test dependencies are required and configured in separate support files to keep the content of the `spec` and `rails` helpers to a minimum.

This change rectifies the issue with simplecov, and implements the pattern, by aligning how rspec is setup to match the WEX service.


Cruikshanks:
@irisfaraway not doing anything about the change in test coverage as that is a result of the fix to simplecov.

PR #280, opened 2019-03-17
Remove passenger from dev group in Gemfile

Cruikshanks:
We added passenger to the dev group in the Gemfile to replace using Webrick as the app server when running the app locally.

The intention was to more closely mirror our setup in production. However on review we have found that there have not been anytimes where this has helped us resolve an issue or develop a new feature. What the dev's often do though is comment it out before running locally so things like **byebug** will work.

So rather than hinder local development for no gain, we have chosen to stop using passenger locally and in our Vagrant build.


Cruikshanks:
Closing, raised in error. This change has already been done.

PR #306, opened 2019-04-29
Remove byebug and swap it with pry-byebug

cintamani:
byebug has a very similar interface as gdb, but byebug does not use the powerful Pry REPL.
binding.pry uses Pry, but lacks some of the byebug features.
[pry-byebug](https://github.com/deivid-rodriguez/pry-byebug) brings some capabilities byebug to binding.pry, so using that, will give you the most debugging powers.

Credits: https://docs.gitlab.com/ee/development/pry_debugging.html#byebug-vs-bindingpry


Gemfile
```diff
@@ -77,8 +77,8 @@ group :production do
 end
 
 group :development, :test do
-  # Call "byebug" anywhere in the code to stop execution and get a debugger console
-  gem "byebug"
+  # Call "buinding.pry" anywhere in the code to stop execution and get a debugger console
```
Cruikshanks:
```suggestion
  # Call "binding.pry" anywhere in the code to stop execution and get a debugger console
```

spec/spec_helper.rb
```diff
@@ -4,8 +4,8 @@
 # This is as per its docs https://github.com/colszowka/simplecov#getting-started
 require "./spec/support/simplecov"
 
-# Support debugging in the tests
-require "byebug"
+# Call "buinding.pry" anywhere in the code to stop execution and get a debugger console
```
Cruikshanks:
```suggestion
# Call "binding.pry" anywhere in the code to stop execution and get a debugger console
```

Gemfile
cintamani: 🤦‍♀ 👓

Issue #206, opened 2018-11-27
Intermittently failing back office user test


irisfaraway:
Occasionally we get this failure:

`rspec ./spec/services/user_migration_service_spec.rb:33 # UserMigrationService#sync when there is a backend agency_user who isn't in the back office creates a new back office user`

```
Failures:
  1) UserMigrationService#sync when there is a backend agency_user who isn't in the back office creates a new back office user
     Failure/Error: expect(User.all.count).to eq(number_of_users_before_sync + 1)
     
       expected: 110
            got: 111
     
       (compared using ==)
     # ./spec/services/user_migration_service_spec.rb:36:in 'block (4 levels) in <top (required)>'
```

Cause is unclear, but due to the intermittency it's likely to depend on the order in which the specs are run.

irisfaraway:
This fails with seed 34601.

PR #322, opened 2019-05-29
Fix database cleaner to keep the test environment clean at every test

cintamani:
Fixes: https://github.com/DEFRA/waste-carriers-back-office/issues/206

The problem was created by the incorrect DB strategy. The DB was cleaned before every suite but not between tests.
Hence, the code in here https://github.com/DEFRA/waste-carriers-back-office/blob/master/spec/services/user_migration_service_spec.rb#L31 was catching the wrong
object.


cintamani:
Oh yes :D I'm on it

PR #340, opened 2019-07-18
Add privacy policy page for renewals

cintamani:
From https://eaflood.atlassian.net/browse/RUBY-302

Add verbal privacy policy screen on the backoffice.
The page's button "continue" will redirect to the beginning of the renewal journey for the given registration.
The front office dashboard renewal link will be changed to point to this page instead of directly the start of the journey.
This is the same strategy used for the verbal privacy policy on WEX.

Screenshot 2019-07-18 at 16 22 36

### DIFF
--- a/app/controllers/ad_privacy_policy_controller.rb
+++ b/app/controllers/ad_privacy_policy_controller.rb
@@ -2,7 +2,6 @@
 
 class AdPrivacyPolicyController < ApplicationController
   before_action :authenticate_user!
-  
 
   def show
     @reg_identifier = params[:reg_identifier]
--- a/app/controllers/ad_privacy_policy_controller.rb
+++ b/app/controllers/ad_privacy_policy_controller.rb
@@ -2,6 +2,7 @@
 
 class AdPrivacyPolicyController < ApplicationController
   before_action :authenticate_user!
+  
 
   def show
     @reg_identifier = params[:reg_identifier]
--- a/app/controllers/ad_privacy_policy_controller.rb
+++ b/app/controllers/ad_privacy_policy_controller.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class AdPrivacyPolicyController < ApplicationController
+  before_action :authenticate_user!
+
   def show
     @reg_identifier = params[:reg_identifier]
   end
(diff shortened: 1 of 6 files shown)


config/routes.rb
```diff
@@ -18,6 +18,9 @@
   get "/bo/convictions/approved" => "convictions_dashboards#approved", as: :convictions_approved
   get "/bo/convictions/rejected" => "convictions_dashboards#rejected", as: :convictions_rejected
 
+  # Privacy policy
+  get "/ad-privacy-policy/:reg_identifier", to: "ad_privacy_policy#show", as: :ad_privacy_policy
```
irisfaraway:
```suggestion
  get "/bo/ad-privacy-policy/:reg_identifier", to: "ad_privacy_policy#show", as: :ad_privacy_policy
```

We have to include /bo at the start of all back office routes. This is because the backend and the back office are on the same subdomain and we need to check the URL to decide which app to route to.

Think this will require an update to https://github.com/DEFRA/waste-carriers-frontend as well.

cintamani: Oh!! will look into that! Locally it worked fine 🤔

irisfaraway: Yeah, locally we have different ports for the different apps so it's not an issue. But on production it would be.

config/locales/ad_privacy_policy.en.yml
```diff
@@ -0,0 +1,40 @@
+en:
+  ad_privacy_policy:
+    show:
+      heading: "Privacy policy for Waste Carriers, Brokers and Dealers assisted digital"
```
irisfaraway:
```suggestion
      title: "Privacy policy for assisted digital"
      heading: "Privacy policy for Waste Carriers, Brokers and Dealers assisted digital"
```

irisfaraway: This will add a unique title attribute to the page. At the moment we just have a generic Waste Carriers title. (Been busted during QA for this before!)

cintamani: Nice one! :D

app/controllers/ad_privacy_policy_controller.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AdPrivacyPolicyController < ApplicationController
+  before_action :authenticate_user!
```
cintamani:
```suggestion
  before_action :authenticate_user!
  
```

app/controllers/ad_privacy_policy_controller.rb
```diff
@@ -0,0 +1,10 @@
...
+
```
cintamani:
```suggestion
```

PR #382, opened 2019-10-11
Include registrations in dashboard search

irisfaraway:

The back office dashboard search was originally designed to only look at renewals - records in the TransientRegistrations table. We want to look at the Registrations table as well so we can see completed registrations and new applications in progress.

This will require updating our search service, as well as changing the design of the dashboard to be able to handle both types of object and show the correct information and actions for them.

### DIFF
--- a/app/views/dashboards/index.html.erb
+++ b/app/views/dashboards/index.html.erb
@@ -79,6 +79,8 @@
                 <% end %>
               </td>
               <td>
+                <% if result.is_a?(WasteCarriersEngine::TransientRegistration) %>
+
                 <% if result_type(result).present? %>
                   <%= t(".results.classes.#{result_type(result)}") %><br>
                 <% end %>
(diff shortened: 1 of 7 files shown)


irisfaraway:
Relies on https://github.com/DEFRA/waste-carriers-engine/pull/511

irisfaraway:
This was starting to get hefty so I've rebased it into several PRs. This does the work of updating the search so that we get the registrations into the results, although they're not displayed the way that we ultimately want them to be.

Once this is merged, we can look at:

- #384 
- #385 
- #386

app/views/dashboards/index.html.erb
```diff
@@ -79,6 +79,8 @@
                 <% end %>
               </td>
               <td>
+                <% if result.is_a?(WasteCarriersEngine::TransientRegistration) %>
+
                 <% if result_type(result).present? %>
```
cintamani: Can we fix indentation? :D

irisfaraway: I think this gets removed by another PR anyway as it was a temporary fix :)

PR #384, opened 2019-10-14
Determine which tags to display in search results

irisfaraway:

This turned out to be quite complex so I created an entire service to deal with it.

Currently this only takes pending payments and convictions for renewals into account - we'll need different logic for new registrations.

### DIFF
--- a/app/services/status_tag_service.rb
+++ b/app/services/status_tag_service.rb
@@ -18,8 +18,18 @@ class StatusTagService < ::WasteCarriersEngine::BaseService
   private
 
   def metadata_status
-    return reg_metadata_status if @resource.is_a?(WasteCarriersEngine::Registration)
-    return transient_reg_metadata_status if @resource.is_a?(WasteCarriersEngine::TransientRegistration)
+    return :revoked if @resource.metaData.REVOKED?
+    return :refused if @resource.metaData.REFUSED?
+
+    if transient?
+      transient_reg_metadata_status
+    else
+      reg_metadata_status
+    end
+  end
+
+  def transient_reg_metadata_status
+    return :in_progress unless @resource.renewal_application_submitted?
   end
 
   def reg_metadata_status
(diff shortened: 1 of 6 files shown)


app/services/status_tag_service.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def pending_payment
+    return nil unless submitted_renewal?
+
+    :pending_payment if @resource.pending_payment?
+  end
+
+  def stuck
+    return nil unless submitted_renewal?
```
Cruikshanks: `stuck?` already checks `renewal_application_submitted?` so I think you could get away with

```suggestion
    return nil if @resource.is_a?(WasteCarriersEngine::Registration)
    
    :stuck if @resource.stuck?
```

app/services/status_tag_service.rb
```diff
@@ -0,0 +1,59 @@
...
+  end
+
+  def reg_metadata_status
+    return :in_progress if @resource.metaData.PENDING?
+
+    @resource.metaData.status.downcase.to_sym
+  end
+
+  def transient_reg_metadata_status
+    return :revoked if @resource.metaData.REVOKED?
```
Cruikshanks: Could the revoked and refused checks be moved to the first 2 lines of `metadata_status` as I think they are applicable to both registrations and transient registrations? Plus if the status is one of these it trumps all others.

app/services/status_tag_service.rb
```diff
@@ -0,0 +1,59 @@
...
+    return :refused if @resource.metaData.REFUSED?
+    return :in_progress unless submitted_renewal?
```
Cruikshanks: I think we can get away with just calling `renewal_application_submitted?` as we already know it's a transient registration here

```suggestion
    return :in_progress unless @resource.renewal_application_submitted?
```

PR #386, opened 2019-10-14
Display 'transfer' link on registration results

irisfaraway:

All registrations should show the transfer link unless they have been revoked or refused, as no more actions are possible then.

### DIFF
--- a/spec/helpers/action_links_helper_spec.rb
+++ b/spec/helpers/action_links_helper_spec.rb
@@ -123,5 +123,23 @@ RSpec.describe ActionLinksHelper, type: :helper do
         end
       end
     end
+
+    describe "#display_transfer_link_for?" do
+      context "when the result is not a Registration" do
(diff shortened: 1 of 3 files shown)


spec/helpers/action_links_helper_spec.rb
```diff
@@ -141,5 +141,23 @@
...
+    describe "#display_transfer_link_for?" do
+      context "when the result is not a Registration" do
+        let(:result) { build(:transient_registration) }
+
+        it "returns false" do
+          expect(helper.display_transfer_link_for?(result)).to eq(false)
+        end
+      end
+
+      context "when the result is a Registration" do
```
Cruikshanks: Would it be worth adding an extra sub-context for refused and revoked registrations?

irisfaraway: Oops, yes, dunno how I overlooked that! I'll get on it 👍

PR #387, opened 2019-10-15
Display correct date on search results

irisfaraway:

All search results should have something in the 'date' column, but what this is and how it is displayed varies based on the result type and its current status.

### DIFF
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -18,4 +18,11 @@ module DashboardsHelper
   def status_tags(result)
     StatusTagService.run(resource: result)
   end
+
+  def result_date(result)
+    I18n.t(".dashboards.index.results.date.started", date: result.metaData
+                                                                 .last_modified
+                                                                 .in_time_zone("London")
+                                                                 .strftime("%d/%m/%Y"))
+  end
 end
--- a/app/helpers/dashboards_helper.rb
+++ b/app/helpers/dashboards_helper.rb
@@ -46,9 +46,9 @@ module DashboardsHelper
   end
 
   def last_modified_date(result)
-    I18n.t(".dashboards.index.results.date.started", date: result.metaData
-                                                                 .last_modified
-                                                                 .in_time_zone("London")
-                                                                 .strftime("%d/%m/%Y"))
+    I18n.t(".dashboards.index.results.date.last_modified", date: result.metaData
+                                                                       .last_modified
+                                                                       .in_time_zone("London")
+                                                                       .strftime("%d/%m/%Y"))
   end
 end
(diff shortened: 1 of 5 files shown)


irisfaraway:
This is probably not the final version but it'll work while we wait for answers about the dates we don't have.

app/helpers/dashboards_helper.rb
```diff
@@ -18,4 +18,37 @@ def result_type(result)
...
+    return expired_date(result) if result.metaData.EXPIRED?
+    return will_expire_date(result) if result.is_a?(WasteCarriersEngine::Registration)
+
+    last_modified_date(result)
+  end
+
+  private
+
+  def refused_or_revoked?(result)
+    %w[REFUSED REVOKED].include?(result.metaData.status)
```
cintamani: I do feel like we are giving helpers knowledge that belongs to the model in regard to accessing the status. Shall we create a `state` method in the models and use that to decide about the status? Even better if we can implement something like:
```
def revoked?
  metaData.status == "REVOKED"
end

def refused?
  metaData.status == "REFUSED"
end
```
In the engine model. 
I would love to have states managed in the same place so that/if they will need update we won't have to change things all over the place. 
In general, hoping for the killing of MongoDB in favour of Postgres, if we can keep every access to `metaData` inside the model and anything else needing it has to pass through a model public interface method, it would be good. 
Specifically, I really would like to see states managed all in one place, proper constant created for the different states and they always accessed through their owner's public interface.

Cruikshanks: I agree with this, though it'd probably need to be a concern applied to both registration and transient_registration (you can refuse a renewal).

However I'd expect it done in a separate PR, not as part of this one. So i'm fine with holding this until that PR is done and merged, or approving this but then submitting the statuses PR's straight after.

irisfaraway: Feast yer eyes on https://github.com/DEFRA/waste-carriers-engine/pull/520

irisfaraway: And fixed: https://github.com/DEFRA/waste-carriers-back-office/pull/387/commits/4228992a86e96b0aedbcbb98757bffa521665bf6

app/helpers/dashboards_helper.rb
```diff
@@ -18,4 +18,38 @@ def result_type(result)
...
+  private
+
+  def deactivated?(result)
+    result.inactive? || result.refused? || result.revoked?
+  end
+
+  def expired_date(result)
+    I18n.t(".dashboards.index.results.date.expired", date: result.expires_on
+                                                                 .in_time_zone("London")
+                                                                 .strftime("%d/%m/%Y"))
```
Cruikshanks: In WEX we've been using the convention of using Rails `time_formats.rb` to centralise our formats https://github.com/DEFRA/waste-exemptions-back-office/blob/master/config/initializers/time_formats.rb

Could we move this there and then switch to using `.to_formatted_s(:day_month_year)))`?

irisfaraway: Good shout. Tidied: https://github.com/DEFRA/waste-carriers-back-office/pull/387/commits/62ead7d9dddba5440b7ac8818f50a0adfa990b7e

PR #390, opened 2019-10-20
Display correct number of results on dashboard

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-696

This fixes a bug found in QA. The result count in the table caption was showing the number on the page, not the total number.

To fix this, we move pagination into the controller and out of the service, which is probably a better fit anyway.

### DIFF
--- a/app/controllers/dashboards_controller.rb
+++ b/app/controllers/dashboards_controller.rb
@@ -7,6 +7,14 @@ class DashboardsController < ApplicationController
 
   def index
     @term = params[:term]
-    @results = SearchService.run(page: params[:page], term: @term)
+    search_and_count_results(params[:page])
+  end
+
+  private
+
+  def search_and_count_results(page)
+    result_data = SearchService.run(page: page, term: @term)
+    @result_count = result_data[:count]
+    @results = result_data[:results]
   end
 end
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -2,21 +2,34 @@
 
 class SearchService < ::WasteCarriersEngine::BaseService
   def run(page:, term:)
-    return [] if term.blank?
+    return response_hash([]) if term.blank?
 
-    @term = term
-    Kaminari.paginate_array(combined_results).page(page)
+    @page = page
+    @term = Regexp.escape(term)
+
+    response_hash(search_results)
   end
 
   private
 
-  def combined_results
-    search(WasteCarriersEngine::TransientRegistration) + search(WasteCarriersEngine::Registration)
+  def response_hash(results)
+    {
+      count: results.count,
+      results: Kaminari.paginate_array(results).page(@page)
+    }
+  end
+
+  def search_results
+    @_search_results ||= matching_resources.sort_by(&:reg_identifier)
+  end
+
+  def matching_resources
+    search(WasteCarriersEngine::Registration) + search(WasteCarriersEngine::TransientRegistration)
   end
 
   def search(model)
     model.search_term(@term)
-         .order_by("metaData.lastModified": :desc)
+         .limit(100)
          .read(mode: :secondary)
   end
 end
--- a/app/services/search_service.rb
+++ b/app/services/search_service.rb
@@ -5,7 +5,7 @@ class SearchService < ::WasteCarriersEngine::BaseService
     return response_hash([]) if term.blank?
 
     @page = page
-    @term = Regexp.escape(term)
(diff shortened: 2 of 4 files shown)


app/controllers/dashboards_controller.rb
```diff
@@ -7,6 +7,20 @@ class DashboardsController < ApplicationController
...
     @term = params[:term]
-    @results = SearchService.run(page: params[:page], term: @term)
+    search_and_count_results(params[:page])
+  end
+
+  private
+
+  def search_and_count_results(page)
+    if @term.present?
+      all_results = SearchService.run(term: @term)
```
Cruikshanks: Hmm 🤔 

I don't believe this change has introduced what may be a performance issue, but I think we may have one.

It looks to me with this line that we are asking Rails to return all matching results from the database, then asking Kaminari to paginate them. I think when we come to use live data and someone puts in a term like `waste` or `limited` this may cause a very slow response, or even crash the page.

So I think we'll want to be doing something similar to what we did in the front-office when dealing with this issue https://github.com/DEFRA/waste-carriers-front-office/blob/master/app/controllers/dashboards_controller.rb#L7-L10

How that affects this is, if we want to show the count we may want to do that separately as a count only i.e. get ActiveRecord/Postgres to only return the count and not the data, rather than count the number of records returned.

irisfaraway: Good point. I'll have a crack at sorting it out!

irisfaraway: Actually, given that we were calling `Kaminari.paginate_array(combined_results).page(page)` previously anyway (just in the service instead) we may have had this issue already. Doesn't mean it's worth fixing now though.

irisfaraway: Looks like this would be a lot easier if we were on Rails 5 and could use `or`: https://stackoverflow.com/a/9540911

Cruikshanks: Lets get this bad boy aligned then upgrade!

irisfaraway:
So it looks like if we want to merge the two ActiveRecord queries in Rails 4 without just creating an array of all the results, we could do one of the following:

- write the query in SQL instead to do a `union` on the two tables
- use this gem to backport the 'or' functionality from Rails 5: https://github.com/Eric-Guo/where-or

Alternatively, we can just fix this as soon as we go to Rails 5. @Cruikshanks @cintamani One to discuss at the dev coffee today?

Cruikshanks:
> So it looks like if we want to merge the two ActiveRecord queries in Rails 4 without just creating an array of all the results, we could do one of the following:
> 
> * write the query in SQL instead to do a `union` on the two tables
> * use this gem to backport the 'or' functionality from Rails 5: https://github.com/Eric-Guo/where-or
> 
> Alternatively, we can just fix this as soon as we go to Rails 5. @Cruikshanks @cintamani One to discuss at the dev coffee today?

I'm not sure the gem, or even the Rails 5 fix is going to provide what you expect. Based on https://github.com/rails/rails/blob/aa3acf85cb/activerecord/test/cases/relation/or_test.rb the `or` is for queries on the same model. The common example is `post = Post.where('id = 1').or(Post.where('id = 2'))`.

A `UNION` on the other hand allows us to combine the results of 2 queries into one, for example your case of querying 2 separate tables to produce one result set.

So I would just go with coding your own SQL query. We should hopefully then be able to pass in `page` and deal with the performance issue.

irisfaraway:
After a lot of shenanigans involving writing SQL queries for a no-SQL database (oops) and then wrestling with Mongoid pipelines, we decided to just limit the results returned so we aren't trying to paginate too much stuff at once. Thanks @Cruikshanks for this idea 🙇‍♀️

Sooo finally ready to review again... as soon as I fix these tests 😭

app/services/search_service.rb
```diff
@@ -2,21 +2,34 @@
...
+  end
+
+  def matching_resources
+    search(WasteCarriersEngine::Registration) + search(WasteCarriersEngine::TransientRegistration)
   end
 
   def search(model)
     model.search_term(@term)
-         .order_by("metaData.lastModified": :desc)
+         .limit(100)
```
cintamani: [NIT] Can I suggest we make this a constant?

irisfaraway: The `100`? I'm not too fussed to be honest, we only use it once and I don't think it'll need to be changed much (if ever).

PR #399, opened 2019-10-23
Display convictions tag for new registrations

irisfaraway:

There's no 'submitted' state for a new registration at the moment, so we will display this tag as soon as the registration is in the database and is known to require a convictions check. So slightly different behaviour from renewals for now but this will be aligned.


irisfaraway:
This is a first pass but decided to do some more of the work in the engine, so will refactor once https://github.com/DEFRA/waste-carriers-engine/pull/535 is merged and rolled out.

PR #400, opened 2019-10-23
Move action link paths to helper

irisfaraway:

This is more in line with WEX. It also means that when we start having different routes for registrations vs transient_registrations, we can handle that in the helper without messing up the view.

### DIFF
--- a/spec/helpers/action_links_helper_spec.rb
+++ b/spec/helpers/action_links_helper_spec.rb
@@ -3,6 +3,114 @@
 require "rails_helper"
 
 RSpec.describe ActionLinksHelper, type: :helper do
+  describe "details_link_for" do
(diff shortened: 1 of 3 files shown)


spec/helpers/action_links_helper_spec.rb
```diff
@@ -3,6 +3,114 @@
...
+    end
+
+    context "when the resource is not a registration" do
+      let(:resource) { create(:transient_registration) }
+
+      it "returns the correct path" do
+        expect(helper.transfer_link_for(resource)).to eq("#")
+      end
+    end
+  end
```
cintamani: Shall we keep this pure unit tests? So the resources are `double` that respond to methods such as `is_a?` with a specific parameter and return a result?

PR #404, opened 2019-10-24
Apply role permissions to dashboard links

irisfaraway:

Agency users vs finance users have different abilities for what tasks they can and cannot do. So we shouldn't display links to actions which aren't possible or relevant.


spec/helpers/action_links_helper_spec.rb
```diff
@@ -60,8 +60,20 @@
       context "when the result is in a resumable state" do
         before { result.workflow_state = "location_form" }
 
-        it "returns true" do
-          expect(helper.display_resume_link_for?(result)).to eq(true)
+        context "when the user does not have permission" do
```
cintamani: Shall we add an integration test scenario for this in our `request` tests?

PR #406, opened 2019-10-25
Make dashboard CSS class names less generic

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-735

This fixes a styling regression bug on the login pages which was causing the background of the 'log in' button to become grey, due to it also having the class 'actions'.


app/views/dashboards/index.html.erb
```diff
@@ -32,7 +32,7 @@
   </div>
 
   <div class="column-one-third">
-    <div class="actions">
+    <div class="dashboard_actions">
```
cintamani:
```suggestion
    <div class="actions">
```

PR #407, opened 2019-10-28
Add call out box for registration info

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-705

Add grey box with registration info in Registration details page

Screenshot 2019-10-28 at 15 16 01
Screenshot 2019-10-28 at 15 14 33

### DIFF
--- a/app/assets/stylesheets/partials/_registrations.scss
+++ b/app/assets/stylesheets/partials/_registrations.scss
@@ -1,4 +1,6 @@
 .panel.wcr-panel{
+  margin-top: 25px;
+  
   &-border-all {
     border: 1px solid #bfc1c3;
   }
--- a/app/assets/stylesheets/partials/_registrations.scss
+++ b/app/assets/stylesheets/partials/_registrations.scss
@@ -1,0 +1,14 @@
+.panel.wcr-panel{
+  &-border-all {
+    border: 1px solid #bfc1c3;
+  }
+
+  p.lede {
+    margin: 0;
+  }
+
+  h2 {
+    margin-bottom: 5px;
+  }
+
+}
--- a/app/views/transient_registrations/show.html.erb
+++ b/app/views/transient_registrations/show.html.erb
@@ -70,6 +70,20 @@
         <% end %>
       </div>
     <% end %>
+
+    <div class="panel wcr-panel wcr-panel-border-all">
+      <h2 class="heading-medium">
+        <%= @transient_registration.company_name %>
+      </h2>
+      <p class="lede">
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/partials/_registrations.scss
```diff
@@ -0,0 +1,14 @@
+.panel.wcr-panel{
```
cintamani:
```suggestion
.panel.wcr-panel{
  margin-top: 25px;
  
```

app/views/transient_registrations/show.html.erb
```diff
@@ -70,6 +70,20 @@
         <% end %>
       </div>
     <% end %>
+
+    <div class="panel wcr-panel wcr-panel-border-all">
+      <h2 class="heading-medium">
+        <%= @transient_registration.company_name %>
+      </h2>
+      <p class="lede">
+        <%= t(".tier.#{@transient_registration.tier.downcase}") %> - <%= show_translation_or_filler(:registration_type) %>
```
Cruikshanks: In the case of lower tier registrations, they never have a `registration_type`.

So if I've understood the code correctly lower tier will always show as **Lower tier - -**. Ideally, for lower tier there wouldn't be any dashes.

app/views/transient_registrations/show.html.erb
```diff
@@ -70,6 +70,20 @@
...
+      </p>
+      <p>
+        <%= t(".labels.expires", formatted_date: display_expiry_date) %>
+        <br>
+        <%= t(".labels.account", email: @transient_registration.account_email) %>
```
Cruikshanks: You might want to use the `show_translation_or_filler()` helper here. It shouldn't happen but seeing as we just ran a data fix today for it, it is possible that a non-AD registration does not have an account email 😩 

```suggestion
        <%= t(".labels.account", email: show_translation_or_filler(:account_email) %>
```

cintamani: No this won't work. that method has expectations on translations, don't just return a value or "-". I think I am going to leave it as it is, if there is no email it will just show `Account:`, no error is thrown, and then we can discuss with Martin if we should drop the line entirely :)

Cruikshanks: Though you lost a bracket when you reverted the change!

Latest is showing as `<%= t(".labels.account", email: @transient_registration.account_email %>`

cintamani: yep! Just added it!! Sorry!

PR #411, opened 2019-10-29
Update conviction section in view details

cintamani:
Closes: https://eaflood.atlassian.net/browse/RUBY-709

Move convictions details section just below business and contact details and change slightly to partially match the wireframes.

Screenshot 2019-10-30 at 16 27 03


app/views/transient_registrations/show.html.erb
```diff
@@ -138,10 +133,32 @@
...
+              <%= link_to t(".conviction_link"), transient_registration_convictions_path(@transient_registration.reg_identifier), class: "pull-up" %>
+            <% end %>
+            <p>
+              <%= t(".conviction_search_text.#{@transient_registration.conviction_check_required?}") %>
+            </p>
+          <% else %>
+            <p>
+              <%= t(".conviction_search_text.unknown") %>
+            </p>
+          <% end %>
```
Cruikshanks: I could be wrong, but I feel this is a good candidate for moving into a helper. From what I can see, we're currently checking various properties of the `transient_registration` to determine 2 things

- should we show a link
- what text to show `t(".conviction_search_text.[KEY]")`

I think now is a good opportunity to move these into helper methods (which we can then unit test) and simplify the view.

```ruby
<% if display_convictions_link?(@transient_registration) %>
  <%= link_to t(".conviction_link"), transient_registration_convictions_path(@transient_registration.reg_identifier), class: "pull-up"  %>
<% end %>
<p>
  <%= convictions_text(@transient_registration) %>
</p>
```

Cruikshanks: That's fine, and with something, as complex as the view details page, we probably need a presenter more than a bunch of helpers.

So agree to bite my tongue and we'll aim to return to this view later.

PR #412, opened 2019-10-29
Add actions box on Registration details page

cintamani:
Closes: https://eaflood.atlassian.net/browse/RUBY-709

Add a grey action box on the RHS column with convictions and payment links.

Screenshot 2019-10-29 at 12 41 55

### DIFF
--- a/app/assets/stylesheets/partials/_registrations.scss
+++ b/app/assets/stylesheets/partials/_registrations.scss
@@ -1,6 +1,6 @@
 .panel.wcr-panel{
   margin-top: 25px;
-  
+
   &-border-all {
     border: 1px solid #bfc1c3;
   }
@@ -12,5 +12,4 @@
   h2 {
     margin-bottom: 5px;
   }
-
 }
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/partials/_registrations.scss
```diff
@@ -0,0 +1,9 @@
+.wcr-action-boxout {
```
Cruikshanks: @irisfaraway PR #377 Also added actions to the dashboard, though she has a different back color (I can see from WEX that we've used [#f8f8f8](https://github.com/DEFRA/waste-exemptions-back-office/blob/master/app/assets/stylesheets/partials/_resource_details.scss#L2) there.

I'm just wondering if there is a way we can define the main class in one place to ensure some consistency, and then build on it as needed in the dashboard and view details page?

cintamani: Oh! I have missed that!! Will reuse that code! I think I did copy the colour and style from WEX entirely. Will also update the colour to match WEX :)

PR #427, opened 2019-11-04
Registration details page - People section

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-751

Implement new design to the transient registration details page to deal with people section

Screenshot 2019-11-04 at 10 39 56

### DIFF
--- a/app/views/transient_registrations/show.html.erb
+++ b/app/views/transient_registrations/show.html.erb
@@ -168,65 +168,6 @@
           <% end %>
         </div>
 
-        <% if @transient_registration.main_people.present? || @transient_registration.relevant_people.present? %>
-          <h2 class="heading-medium"><%= t(".people_heading") %></h2>
-        <% end %>
-
-        <% if @transient_registration.main_people.present? %>
-          <% @transient_registration.main_people.each do |person| %>
-            <table>
-              <caption class="heading-small">
-                <%= t(".person_information.heading.#{@transient_registration.business_type}") %>
-              </caption>
-              <tbody>
-                <%= render("shared/person_table_rows", person: person) %>
-                <tr>
-                  <th>
-                    <%= t(".person_information.labels.conviction_search_result") %>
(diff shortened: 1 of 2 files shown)


app/views/transient_registrations/show.html.erb
```diff
@@ -302,6 +243,48 @@
...
+              <% @transient_registration.main_people.each do |person| %>
+                <tr>
+                  <td>
+                    <%= person.first_name %> <%= person.last_name %>
+                  </td>
+                  <td>
+                    <%= person.dob.to_date %>
+                  </td>
+                  <td>
+                    <% if person.conviction_search_result.present? %>
```
irisfaraway: I think all people will have a `conviction_search_result` - it's whether the `match_result` is positive or negative that we want to check.

The only exception is if the application is still being filled in and the check hasn't run yet.

app/views/transient_registrations/show.html.erb
```diff
@@ -302,6 +243,48 @@
...
+                      <div class="bold-xsmall">
+                        <%= t(".person_information.has_convictions.yes") %>
+                      </div>
+                    <% else %>
+                      <%= t(".person_information.has_convictions.no") %>
```
irisfaraway: Why is the `yes` styled to be small and the `no` isn't?

app/views/transient_registrations/show.html.erb
cintamani: Oh I see!! Thanks! :)

app/views/transient_registrations/show.html.erb
cintamani: is not small, is `bold` :D GDS Elements

app/views/transient_registrations/show.html.erb
cintamani:
```suggestion
                    <% if person.conviction_check_required? %>
```

PR #439, opened 2019-11-12
Refactor UserMigrationService to inherit from BaseService

irisfaraway:
We want to refactor our services to all follow a similar format. This means we can now just call UserMigrationService.run instead of instatiating a new service and then calling several methods on it.

Plus minor tweaks to the UserMigrationsController which calls it.

### DIFF
--- a/spec/services/user_migration_service_spec.rb
+++ b/spec/services/user_migration_service_spec.rb
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 RSpec.describe UserMigrationService do
-  subject do
+  let(:service) do
     UserMigrationService.run
   end
 
@@ -15,7 +15,7 @@ RSpec.describe UserMigrationService do
     context "when there are no backend users" do
       it "does not modify the back office users" do
         users_before_sync = User.all
-        subject
+        service
         expect(User.all).to eq(users_before_sync)
       end
     end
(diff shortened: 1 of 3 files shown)


spec/services/user_migration_service_spec.rb
```diff
@@ -3,25 +3,19 @@
 require "rails_helper"
 
 RSpec.describe UserMigrationService do
-  let(:user_migration_service) do
-    UserMigrationService.new
+  subject do
+    UserMigrationService.run
```
cintamani: I am not sure I like using `subject` here, one for discussion :D

irisfaraway: @cintamani Swapped it for the format we used on the other services – how's this? https://github.com/DEFRA/waste-carriers-back-office/pull/439/commits/b8e18795f17049cfcbfde094472bae36c14ccf59

cintamani: Nice!! What if we call it `result` rather than `service`? IMO seems clearer what the variable holds with `result`

Cruikshanks: My NIT for this is I don't think we should bother with this let at all. That might be because I don't understand how it's helping.

It would be clearer to me if I just saw `UserMigrationService.run` instead of `service` in the tests, or `described_class.run` to avoid lots of changes should we rename it.

But as I said, there might be a benefit or convention of putting the execution in a `let` I'm just not aware of.

cintamani: I think the benefit of having it in a `let` is that if the execution of the service changes we have to update 1 line of code only

Cruikshanks: The other PR helped me see the light https://github.com/DEFRA/waste-carriers-back-office/pull/440#pullrequestreview-316107268 !

Cruikshanks: Perhaps `run_service` 🤷‍♂

PR #440, opened 2019-11-12
Refactor RegistrationTransferService to inherit from BaseService

irisfaraway:
We want to refactor our services to all follow a similar format. This means we can now just call `RegistrationTransferService.run` instead of instantiating a new service and then calling several methods on it.

Also some minor tweaks to the RegistrationTransferForm which calls it.

### DIFF
--- a/spec/services/registration_transfer_service_spec.rb
+++ b/spec/services/registration_transfer_service_spec.rb
@@ -3,68 +3,51 @@
 require "rails_helper"
 
 RSpec.describe RegistrationTransferService do
-  let(:transient_registration) do
-    create(:renewing_registration, :ready_to_renew)
+  let(:registration) { create(:registration) }
+  let(:recipient_email) { external_user.email }
+
+  let(:service) do
+    RegistrationTransferService.run(registration: registration, email: recipient_email)
   end
 
-  let(:registration) do
-    WasteCarriersEngine::Registration.where(reg_identifier: transient_registration.reg_identifier).first
-  end
-
-  let(:registration_transfer_service) do
-    RegistrationTransferService.new(registration)
-  end
-
-  describe "#initialize" do
-    context "when a transient_registration exists" do
-      it "sets @transient_registration" do
-        instance_var = registration_transfer_service.instance_variable_get(:@transient_registration)
-        expect(instance_var).to eq(transient_registration)
-      end
-    end
-
-    context "when no transient_registration exists" do
-      let(:registration_transfer_service) do
-        RegistrationTransferService.new(create(:registration))
-      end
-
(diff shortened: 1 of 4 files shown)


spec/services/registration_transfer_service_spec.rb
```diff
@@ -3,68 +3,51 @@
...
-        RegistrationTransferService.new(create(:registration))
-      end
-
-      it "sets @transient_registration to nil" do
-        instance_var = registration_transfer_service.instance_variable_get(:@transient_registration)
-        expect(instance_var).to eq(nil)
-      end
-    end
+  let(:service) do
+    RegistrationTransferService.run(registration: registration, email: recipient_email)
```
Cruikshanks: I guess this is a possible answer to https://github.com/DEFRA/waste-carriers-back-office/pull/439#discussion_r345645278 !

By putting it in a let we avoid duplicating `described_class.run(registration: registration, email: recipient_email)` in each test, and possibly needing to make lots of updates if we change the method signature.

Think I'd still prefer to just see the call made in the tests, but that's very much a personal opinion. If we stick with the convention though I think I'd see if we could come up with a more expressive name. I know @cintamani suggested `:result` but I think I would opt for something like `:execute_service`.

Not all our services return a result from `run()`, and if this becomes a convention in our service specs it wouldn't always fit. Plus where we're not testing the result it doesn't look quite right

```ruby
old_emails_sent_count = ActionMailer::Base.deliveries.count
result
expect(ActionMailer::Base.deliveries.count).to eq(old_emails_sent_count + 1)

# ...

old_emails_sent_count = ActionMailer::Base.deliveries.count
execute_service
expect(ActionMailer::Base.deliveries.count).to eq(old_emails_sent_count + 1)
```

Should we be testing the result of `run()` I'd then probably do the following to make it clear

```ruby
result = execute_service
expect(result).to eq(:success_existing_user)
```

cintamani: I think whatever works for you is ok. I would argue that `execute_service` reminds me of a method rather than a variable, but really this is only about let's decide what we find clearer and go with it :) I am used to the `result` let variable in services as that is a convention that existed in almost all my previous jobs so I am just used to that. This will work too

Cruikshanks: > reminds me of a method rather than a variable

I think that's why I flagged this in the first place. In a lot of the tests, we don't care what the value of `service` is, it's just an alias for calling the `run()` method.

So I'm happy to use the convention for the reasons you highlighted, but in the tests, I think I'd want to make it clearer the intent is more to _do something_ when we call this `let` rather than simply grabbing a variable.

PR #445, opened 2019-11-14
Add registration details page

cintamani:
This PR contains all the necessary code for showing a details page for registration. 
A presenter has been used instead of a helper in the view. This is because we have already previously decided that we wanted a presenter for the details page as the business logic is very complex. I had to cut refactoring somewhere, hence the presenter will need some further work to move more of the view logic into it. 
There are still some outstanding questions regarding: 
 - Show `in progress` / `active` status somewhere. Will need to speak with Martin. 
 - I don't know which page the conviction checks link should redirect to. Will need to speak with Alan. 

Below some screenshots of the page for registration in different states: 

Screenshot 2019-11-15 at 14 24 35
Screenshot 2019-11-15 at 14 24 30
Screenshot 2019-11-15 at 13 36 09
Screenshot 2019-11-15 at 13 36 04
Screenshot 2019-11-15 at 13 35 06
Screenshot 2019-11-15 at 13 34 25
Screenshot 2019-11-15 at 13 34 18
Screenshot 2019-11-15 at 13 34 08


cintamani:
Hakiri is a won't fix. Codeclimate is a: "leave it there", will fix when we refactor renewing registration details page to use a presenter and consolidate the logic of the 2 in separate shared bits

app/views/registrations/show.html.erb
```diff
@@ -0,0 +1,274 @@
...
+
+          <% if @registration.pending_payment? %>
+            <div class="panel">
+              <h2 class="heading-medium">
+                <%= t(".status.headings.pending_payment") %>
+              </h2>
+              <p>
+                <%= t(".status.messages.pending_payment", total: display_pence_as_pounds(@registration.finance_details.balance)) %>
+              </p>
+              <%= link_to t(".status.actions.payment_button"), @registration.finance_details_link, class: 'button' %>
```
Cruikshanks: Do we know why Hakiri (more likely brakeman under the covers) is seeing this line as

```ruby
link_to(t(".status.actions.payment_button"), RegistrationPresenter.new(WasteCarriersEngine::Registration.find_by(:reg_identifier => params[:reg_identifier]), view_context).finance_details_link, :class => "button")
```

I can see we get the registration in the controller using `params[:reg_identifier]` but after that everything we refer to in the presenter comes from the database. So I'd agree it's a "won't fix - what are you on about Hakiri!"

Unfortunately, I've just been in a classroom with [Ryan Ackroyd](https://en.wikipedia.org/wiki/Ryan_Ackroyd) for 2 days so now I'm a little paranoid about everything!

spec/presenters/registration_presenter_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  end
+
+  describe "#finance_details_link" do
+    let(:registration) { double(:registration, id: "12345") }
+
+    before do
+      expect(Rails.configuration).to receive(:wcrs_frontend_url).and_return("http://example.com")
+    end
+
+    it "returns a link to the finance details page in thee old system" do
```
Cruikshanks:
```suggestion
    it "returns a link to the finance details page in the old system" do
```

app/views/registrations/show.html.erb
cintamani: I think Hakiri is complaining about the fact that the link will go to a different domain than the application but I am not super sure 🤔

cintamani:
@Cruikshanks can you also let me know what link for `conviction` checks is correct? I couldn't find the page itself in the old code :/

PR #447, opened 2019-11-18
Add Unknown status to conviction cheks view details

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-751

Add `UNKNOWN` specific status on registrations detail pages

### DIFF
--- a/config/locales/registrations.en.yml
+++ b/config/locales/registrations.en.yml
@@ -59,7 +59,7 @@ en:
         convictions: "Convictions"
         has_convictions:
           "YES": "Yes"
-          "NO": "No"
+          "no": "No"
           "UNKNOWN": "Unknown"
         heading:
           localAuthority: "Chief executive details"
--- a/config/locales/registrations.en.yml
+++ b/config/locales/registrations.en.yml
@@ -58,8 +58,9 @@ en:
         dob: "Date of birth"
         convictions: "Convictions"
         has_convictions:
-          "yes": "Yes"
-          "no": "No"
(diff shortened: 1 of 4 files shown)


config/locales/registrations.en.yml
```diff
@@ -58,8 +58,9 @@ en:
         dob: "Date of birth"
         convictions: "Convictions"
         has_convictions:
-          "yes": "Yes"
+          "YES": "Yes"
           "no": "No"
```
irisfaraway:
```suggestion
          "NO": "No"
```

config/locales/renewing_registrations.en.yml
```diff
@@ -58,8 +58,9 @@ en:
         dob: "Date of birth"
         convictions: "Convictions"
         has_convictions:
-          "yes": "Yes"
+          "YES": "Yes"
           "no": "No"
```
irisfaraway:
```suggestion
          "NO": "No"
```

cintamani: Nope! :) We show `no` even when there is no result, so in the case of `no` it will be downcase. Only `YES` and `UNKNOWN` will happen to be called there

irisfaraway: I see now 🤦‍♀ Ignore me then!

cintamani: If you look at my commits, you will see I have made the same error and had to fix it!! :D

PR #453, opened 2019-11-19
Create presenter for RenewingRegistration details page

cintamani:
This PR creates a presenter for RenewingRegistration, as we have done for Registration in details page.
It also includes a BaseRegistrationPresenter, which will take care of implementing shared functionality between the views.
Next step is adding partials for the different sections so that we can share code in the views which are incredibly long.


app/presenters/renewing_registration_presenter.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+class RenewingRegistrationPresenter < BaseRegistrationPresenter
+  def display_current_workflow_state
+    "#{I18n.t(".renewing_registrations.show.status.messages.in_progress")} \"#{current_workflow_state}\""
```
cintamani:
```suggestion
    "#{I18n.t('.renewing_registrations.show.status.messages.in_progress')} \"#{current_workflow_state}\""
```

cintamani:
@irisfaraway can you push the button? :) 
@Cruikshanks can you mark the false positive in Hakiri?

PR #458, opened 2019-11-20
Add renew link to actions panel in details page

cintamani:
from: https://eaflood.atlassian.net/browse/RUBY-723

Add renew link to registration details page

### DIFF
--- a/app/views/shared/registrations/_action_links_panel.html.erb
+++ b/app/views/shared/registrations/_action_links_panel.html.erb
@@ -4,7 +4,7 @@
   </h2>
   <ul>
     <!-- TODO: This `can?` will need to be updated when we build NewRegistration -->
-    <% if can?(:renew, resource) && resource.in_progress? %>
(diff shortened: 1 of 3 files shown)


app/views/shared/registrations/_action_links_panel.html.erb
```diff
@@ -4,7 +4,7 @@
   </h2>
   <ul>
     <!-- TODO: This `can?` will need to be updated when we build NewRegistration -->
```
Cruikshanks: Can we delete this TODO now?

cintamani: Not really 🤔 We still don't have a `NewRegistration` 🤷‍♀

Cruikshanks: Ok, can we update it then?

```suggestion
    <!-- TODO: This `display_resume_link_for?` will need to be updated when we build NewRegistration -->
```

PR #459, opened 2019-11-21
Display 'expired' instead of 'expires' if expired

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-755

This updates the registration and transient_registration details pages to say 'expired' instead of 'expires' when the resource has already expired.


irisfaraway:
@Cruikshanks Think you're right. I'll have a bash at it!

PR #460, opened 2019-11-21
Add ceased/revoked info box to view details pages

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-754

If a resource has been ceased or revoked, we want to display some information about this on the 'view details' page.

This should include whether it is considered ceased or revoked, who did it and when, and what the reason was. The reason should only be visible for certain user types.


app/views/shared/registrations/_ceased_revoked_panel.html.erb
```diff
@@ -0,0 +1,16 @@
+<% if resource.revoked? || resource.inactive? %>
```
Cruikshanks: I'm going start sounding like a stuck record soon 😳 !

Do you think some of the logic in this partial are candidates for being moved to the `RegistrationPresenter`?

Specifically, I'm thinking of

```ruby
<% if resource.revoked? || resource.inactive? %>
```
```ruby
      <% if resource.revoked? %>
        <%= t(".heading.revoked") %>
      <% elsif resource.inactive? %>
        <%= t(".heading.ceased") %>
      <% end %>
```
```ruby
<% if resource.metaData.revoked_reason.present? && can?(:view_revoked_reasons, resource) %>
```

irisfaraway: I actually started putting some of them in the presenter but it seemed like I was just creating more code for things that were fairly simple and readable as-is. Maybe we need to agree a principle of when things go in the presenter and when they don't?

Being able to check abilities in the presenter will also require us to make some changes to the ability model – nothing major but it is an extra thing.

Cruikshanks: Don't worry about the last one, but I would push for the others.

IMHO moving things to a presenter

- makes them easier to test
- makes them _more_ readable ('cause we lose the ERB cruft)
- makes the view clearer as its more about the structure and less about code

My general principle is if its anything more than an `if foo do bar` I'd start thinking about a presenter. But you are right, we need to agree a team principle for this. Who wants ☕️ and 🍰 later!?

irisfaraway: @Cruikshanks Sounds good to me! 👍

irisfaraway:
Resolved a merge conflict and it dismissed the review! 😢 Sorry @Cruikshanks, could you approve again?

PR #462, opened 2019-11-21
Add all action links to details page

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-723

Add remaining links to the details page for both renewing registrations and registrations.

Screenshot 2019-11-21 at 11 11 51
Screenshot 2019-11-21 at 11 20 21

### DIFF
--- a/app/views/shared/registrations/_finance_details.html.erb
+++ b/app/views/shared/registrations/_finance_details.html.erb
@@ -10,7 +10,7 @@
 
 <% if resource.show_finance_details_link? %>
   <p>
-    <%= link_to t(".finance_information.section_link_label"), payment_link_for(resource) %>
+    <%= link_to t(".finance_information.section_link_label"), resource.finance_details_link %>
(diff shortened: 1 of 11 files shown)


app/views/shared/registrations/_finance_details.html.erb
```diff
@@ -10,7 +10,7 @@
 
 <% if resource.show_finance_details_link? %>
   <p>
-    <%= link_to t(".finance_information.section_link_label"), payment_link_for(resource) %>
+    <%= link_to t(".finance_information.section_link_label"), resource.finance_details_link %>
```
cintamani: @irisfaraway @Cruikshanks This is a good example of Helper vs Presenter, we were discussing before. 
I thought about it, and I think that we should move the code out of the helper if we decide is worth / we have the time to create a presenter for the dashboard results as well. Otherwise, we can live on enough with this as a mixed solution.

PR #465, opened 2019-11-21
Move shared panels to partials on details page views

cintamani:
As part of the work with Details page we want to make sure we have partial containin the logic of the shared panels between the registrations.

### DIFF
--- a/app/presenters/registration_presenter.rb
+++ b/app/presenters/registration_presenter.rb
@@ -1,12 +1,6 @@
 # frozen_string_literal: true
 
 class RegistrationPresenter < BaseRegistrationPresenter
-  def displayable_location
-    location = show_translation_or_filler(:location)
-
-    I18n.t(".registrations.show.business_information.labels.location", location: location)
-  end
-
   def rejected_header
     I18n.t(".registrations.show.status.headings.rejected")
   end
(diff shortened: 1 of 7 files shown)


app/presenters/registration_presenter.rb
```diff
@@ -1,12 +1,6 @@
 # frozen_string_literal: true
 
 class RegistrationPresenter < BaseRegistrationPresenter
-  def displayable_location
-    location = show_translation_or_filler(:location)
-
-    I18n.t(".registrations.show.business_information.labels.location", location: location)
-  end
```
cintamani: This is a left-over of previous development, I have implemented this method in the base presenter already, so we don't need this version any more, and with the shared partial it was causing errors for missing translations hence I had to delete it here.

config/locales/registrations.en.yml
```diff
@@ -27,6 +27,11 @@ en:
           convictions_check_button: "Check conviction matches"
           revert_to_payment_summary_button: "Send back to payment summary"
           missed_worldpay_payment_button: "Add a missed WorldPay payment"
+      tier:
```
irisfaraway: This is in config/locales/shared/registrations.en.yml as well now – should this have been added here? (Same goes for renewing_registrations.en.yml)

cintamani: Nope :/ conflicts madness XD

irisfaraway:
@cintamani Looks like the contact and business info yaml is still in the original files as well as in the new partial one - can you pull that out as well? (We continue to be in merge hell!)

PR #467, opened 2019-11-22
Refactor ability specs

irisfaraway:
This PR will refactor our ability specs to bring them more into line with WEX. The intent is to make them more reusable and easier to read. To do this we'll use more shared examples and separate ability testing out of our users_spec.rb.


irisfaraway:
Still haven't managed to get doubles working for the resources that we check abilities against (eg Registrations). But I'm planning to make some changes to the abilities themselves which may mean we no longer need this anyway. So planning to go with this for now and revisit later.

PR #470, opened 2019-11-25
Make 'revert to payment summary' button display

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-781

This button was never showing because CanCanCan was expecting a RenewingRegistration but instead received a RenewingRegistrationPresenter.

We don't really need to check the class anyway so we can just change the permissions to accept `:all` instead.


cintamani:
Seems the best solution! 💯

PR #476, opened 2019-11-26
Redirects after a renewing is confirmed

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-782

When confirming a payment or approving convictions, a renewing registration might get confirmed and evolve to a registration. In that case, the details page of the renewing registration will not exist any more. Historically, we used to just redirect the user to the main dashboard, but during one of the many refactoring we have killed that functionality, and the test suite didn't highlight the issue due to having different mongoId settings than other environments. Hence, I have also killed those special settings as part of this story.
Now, if a user gets to the renewal details page with a renewal id that no longer exist, we will check if we have a registration matching that id and we will redierct the user there. If there is not even a registration matching the given id, we will redirect to the dashboard as we were historically doing.

### DIFF
--- a/app/controllers/renewing_registrations_controller.rb
+++ b/app/controllers/renewing_registrations_controller.rb
@@ -7,7 +7,7 @@ class RenewingRegistrationsController < ApplicationController
     begin
       transient_registration = WasteCarriersEngine::RenewingRegistration.find_by reg_identifier: params[:reg_identifier]
     rescue Mongoid::Errors::DocumentNotFound
-      scope = WasteCarriersEngine::Registration.where(reg_identifier: params[:reg_identifier]).any?
+      scope = WasteCarriersEngine::Registration.where(reg_identifier: params[:reg_identifier])
 
       return redirect_to registration_path(params[:reg_identifier]) if scope.any?
 
--- a/app/controllers/renewing_registrations_controller.rb
+++ b/app/controllers/renewing_registrations_controller.rb
@@ -5,13 +5,13 @@ class RenewingRegistrationsController < ApplicationController
 
   def show
     begin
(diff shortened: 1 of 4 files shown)


spec/requests/renewing_registrations_spec.rb
```diff
@@ -33,8 +33,19 @@
       end
 
       context "when no matching transient_registration exists" do
+        context "when a registration exist with that identifier" do
```
irisfaraway:
```suggestion
        context "when a registrations exist with that reg_identifier" do
```

spec/requests/renewing_registrations_spec.rb
```diff
@@ -33,8 +33,19 @@
...
+          let(:registration) { create(:registration) }
+
+          it "rerdirects to the registration details page" do
```
irisfaraway:
```suggestion
          it "redirects to the registration details page" do
```

app/controllers/renewing_registrations_controller.rb
```diff
@@ -4,9 +4,15 @@ class RenewingRegistrationsController < ApplicationController
   before_action :authenticate_user!
 
   def show
-    transient_registration = WasteCarriersEngine::RenewingRegistration.find_by(reg_identifier: params[:reg_identifier])
+    begin
+      transient_registration = WasteCarriersEngine::RenewingRegistration.find_by reg_identifier: params[:reg_identifier]
```
irisfaraway:
```suggestion
      transient_registration = WasteCarriersEngine::RenewingRegistration.find_by(reg_identifier: params[:reg_identifier])
```

Can we put brackets around this for consistency? 🙏  I know it makes the line too long and annoys Rubocop but I'd rather find another way around it.

cintamani: Sure :) 

_she hates rubocop_

PR #478, opened 2019-11-27
Remove all links to old backend

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-786

Since the two system do not share login cookies, when moving from the new back office to the old back office users are asked to login again every time. In the case of starting a new registration, they were blocked to do so half way through the process and redirected to login and then to the admin's dashboard.
Since there is currently no functionality on the old code to deal with the redirect to the correct page after login, we decided to kill all the links for our initial release.

### DIFF
--- a/app/views/dashboards/index.html.erb
+++ b/app/views/dashboards/index.html.erb
@@ -31,19 +31,19 @@
     </div>
   </div>
 
-  <div class="column-one-third">
+  <!-- TODO: restore link when internal route exists https://eaflood.atlassian.net/browse/RUBY-786 -->
+  <!-- <div class="column-one-third">
     <div class="wcr-actions">
       <h2 class="heading-small">
-        <%= t(".actions.subheading") %>
+        <%#= t(".actions.subheading") %>
       </h2>
       <ul>
         <li>
-          <!-- TODO: restore link when internal route exists https://eaflood.atlassian.net/browse/RUBY-786 -->
           <%#= link_to t(".actions.new_registration"), "#" %>
         </li>
       </ul>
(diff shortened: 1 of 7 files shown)


cintamani:
Ignore codeclimate

app/views/dashboards/index.html.erb
```diff
@@ -38,8 +38,8 @@
       </h2>
       <ul>
         <li>
-          <%= link_to t(".actions.new_registration"),
-                      "#{Rails.configuration.wcrs_backend_url}/registrations/start" %>
+          <!-- TODO: restore link when internal route exists https://eaflood.atlassian.net/browse/RUBY-786 -->
```
irisfaraway: Can we comment out the whole `div` instead? This just leaves it as a heading and nothing else.

app/views/shared/registrations/_action_links_panel.html.erb
```diff
@@ -30,7 +30,8 @@
       </li>
     <% end %>
 
-    <% if display_view_confirmation_letter_link_for?(resource) %>
+    <!-- TODO: put back the check when internal route exists https://eaflood.atlassian.net/browse/RUBY-786 -->
+    <% if false #display_view_confirmation_letter_link_for?(resource) %>
```
irisfaraway: Why not modify `display_view_confirmation_letter_link_for?(resource)` instead? It would be more consistent with the others.

cintamani: yep!! :) Thanks for catching this

PR #479, opened 2019-11-27
Add classes for reports exports

cintamani:
This PR contains code that will be needed to deal with the exporting tickets that are lined up for next sprint.
It includes the `reports` classes, adds a schedule file and the `whenever` gem, adds defra-ruby-aws gem and configs for EPR export.


cintamani:
Sorry, ignore this. I do need to add more test coverage!

PR #482, opened 2019-12-02
Redirect to registration details after transfering

cintamani:
Since now we have an internal details page that works with registrations, we want users transferring registration to be redirected to it after completing a transfer.

### DIFF
--- a/app/views/registration_transfers/success.html.erb
+++ b/app/views/registration_transfers/success.html.erb
@@ -19,7 +19,7 @@
     </div>
 
     <div class="form-group">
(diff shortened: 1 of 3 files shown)


app/views/registration_transfers/success.html.erb
```diff
@@ -19,7 +19,7 @@
     </div>
 
     <div class="form-group">
-      <%= link_to t(".button"), continue_after_transfer_link(@registration), class: "button" %>
```
irisfaraway: Does this mean we can remove `continue_after_transfer_link` from RegistrationTransfersHelper?

cintamani: Yes! Good catch :)

PR #483, opened 2019-12-02
Update conviction dashboard to include registrations

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-791

Currently, the screen only handles submitted renewals which are awaiting approval.

In the old backend, there is no such screen. Users have to use BOXI to identify those with outstanding checks, then use the dashboard search to find and process them.

As part of the migration to the back-office, NCCC has asked that this screen be the only mechanism for accessing convictions checks in order to help streamline the process. So this PR updates the dashboard to include registrations which require approval as well.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DEFRA/waste-carriers-engine
-  revision: af80154a8e0e6e6d528bc7434ce31e31171ccf6d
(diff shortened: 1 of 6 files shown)


irisfaraway:
Note that if you run this as-is and make a new reg which triggers a conviction check, it won't be listed as that `conviction_sign_off` has no `workflow_state` for us to check.

Decided to tackle this issue separately, so this PR should be ready for review.

PR #486, opened 2019-12-03
Update conviction details page to work with new registrations

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-792

This PR will set up the conviction details page to work with registrations as well as renewals.

It covers setting up the routes, controllers and views so the page loads correctly.

It does not include any of the actions you can take from here.


irisfaraway:
I think a lot of the ConvictionsHelper can be moved into the presenters but I'd like to tackle that in a separate PR to avoid this one getting too bloated.

PR #497, opened 2019-12-05
Implement revocal or refusal of reg after conviction check

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-794

This PR implements the changes from DEFRA/waste-carriers-engine#581 and makes a few tweaks to get it working.

---

Depends on DEFRA/waste-carriers-engine#581


irisfaraway:
There's quite a lot broken here but I think that may be down to https://github.com/DEFRA/waste-carriers-engine/pull/577! So we need to fix that separately.

PR #499, opened 2019-12-06
Complete registration if possible after conviction approval

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-793

When a new registration's convictions are approved and the balance is paid, we can now call the new RegistrationCompletionService from https://github.com/DEFRA/waste-carriers-engine/pull/583

This will mark the registration as active.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DEFRA/waste-carriers-engine
-  revision: 66e06b533ec6c015c58a50779385b903995db40b
+  revision: de2c090f1a12fb3fc59aebe1e2e68606743ccb04
   branch: master
   specs:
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DEFRA/waste-carriers-engine
-  revision: 66e06b533ec6c015c58a50779385b903995db40b
+  revision: de2c090f1a12fb3fc59aebe1e2e68606743ccb04
```
cintamani: Shall we leave this to dependabot?

Cruikshanks: I think I go the other way. If a change in a host app requires an updated version of the engine, I'd rather the update is done with the change, rather than separately. This way its clearer to me that change depended on that version of the engine.

PR #507, opened 2019-12-13
Fix assisted digital continue link

irisfaraway:
Following https://github.com/DEFRA/waste-carriers-back-office/pull/502 we noticed that this continue link didn't always work depending on whether the registration already had a renewal in progress or not.

The new behaviour is that if a renewal is in progress, we will use the token for it. If not, we use the reg_identifier, which is acceptable when no token exists.

This will probably need further changes if the page is ever required for other journeys (eg editing).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/DEFRA/waste-carriers-engine
-  revision: 1c270a93698adfd0c4f0fd215d011ea7ad0a8845
+  revision: 2d3f0995f8bd0787f47f3374eddb4c09671fa96b
   branch: master
   specs:
     waste_carriers_engine (0.0.1)


app/controllers/ad_privacy_policy_controller.rb
```diff
@@ -4,6 +4,19 @@ class AdPrivacyPolicyController < ApplicationController
...
+  private
+
+  def continue_link
+    reg_identifier = params[:reg_identifier]
+    renewing_registration = WasteCarriersEngine::RenewingRegistration.where(reg_identifier: reg_identifier).first
+
+    if renewing_registration.present?
+      WasteCarriersEngine::Engine.routes.url_helpers.new_renewal_start_form_path(token: renewing_registration.token)
+    else
+      WasteCarriersEngine::Engine.routes.url_helpers.new_renewal_start_form_path(token: reg_identifier)
```
cintamani: We have code in place that will redirect to the correct journey page for renewals whenever you hit the `renew` start page with a reg_identifier, so we don't need this at all :) just always send the `token` to the new renewal start form path as a `token`

irisfaraway: That was what I tried at first but it didn't work! I'll try again and see if I overlooked something obvious.

cintamani: Oh! We have merged that fix yesterday I think. If it doesn't work anyway, let me know as that should work, so I'll fix it :)

irisfaraway: Ahhhhhhhhhh. That explains it. Will update the engine gem!

irisfaraway: Yes, that solved it. Will rework the PR.

cintamani: Cool stuff :)

irisfaraway:
Actually it's purely a gem update now so I'm just going to kick dependabot.

PR #513, opened 2019-12-16
Update user list

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-759

Currently, the back office has users and a screen which lists them. However it's just a read-only view. As we move more user functionality in the back office, we need to update the user dashboard to include them.

This PR adds additional columns and links for the user status and actions.

### DIFF
--- a/app/helpers/users_helper.rb
+++ b/app/helpers/users_helper.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module UsersHelper
+  def display_user_actions?(displayed_user, current_user)
+    agency_with_permission?(displayed_user, current_user) || finance_with_permission?(displayed_user, current_user)
+  end
+
+  private
+
+  def agency_with_permission?(displayed_user, current_user)
+    displayed_user.in_agency_group? && current_user.can?(:manage_agency_users, displayed_user)
+  end
+
+  def finance_with_permission?(displayed_user, current_user)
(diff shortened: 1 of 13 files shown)


app/helpers/users_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module UsersHelper
```
Cruikshanks: ❤️ 😍

app/models/user.rb
```diff
@@ -31,6 +31,22 @@ def invalidate_all_sessions!
...
+    %w[finance finance_admin finance_super].include?(role)
+  end
+
+  # Permissions
+
+  def ability
+    @ability ||= Ability.new(self)
+  end
+
+  delegate :can?, :cannot?, to: :ability
```
cintamani: [NIT] Can we move this bit to the top of the class with the rest of the class methods?

app/models/user.rb
```diff
@@ -31,6 +31,22 @@ def invalidate_all_sessions!
...
+  end
+
+  def in_finance_group?
+    %w[finance finance_admin finance_super].include?(role)
+  end
+
+  # Permissions
+
+  def ability
+    @ability ||= Ability.new(self)
```
cintamani: [NIT]
```suggestion
    @_ability ||= Ability.new(self)
```

PR #514, opened 2019-12-16
Activate and deactivate back office users

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-762

Super users should have the ability to activate and deactivate accounts of their group (eg agency supers can deactivate agency accounts, but not finance accounts).

If a user logs into a deactivated account, they should see a message telling them the account is disabled. They should not be able to access any information or make any changes. Active accounts can access and use the system as normal.

This means we can deal with revoking access to the system without entirely deleting all record of the user (which will be important for version control, etc later on).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,7 +40,7 @@ class ApplicationController < ActionController::Base
     return if skip_auth_on_this_controller?
 
     authenticate_user!
-    redirect_to "/bo/pages/deactivated" if current_user_cannot_use_back_office?
+    redirect_to "/bo/pages/deactivated" if current_user_is_deactivated?
   end
 
   # http://jacopretorius.net/2014/01/force-page-to-reload-on-browser-back-in-rails.html
@@ -59,7 +59,7 @@ class ApplicationController < ActionController::Base
     controller.include?("pages") || controller.include?("devise")
   end
 
-  def current_user_cannot_use_back_office?
+  def current_user_is_deactivated?
     # Don't try to check user permissions if the user isn't logged in
     return false if current_user.blank?
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
+  before_action :authenticate_and_authorize_active_user
   before_action :back_button_cache_buster
 
   helper WasteCarriersEngine::ApplicationHelper
@@ -35,10 +36,33 @@ class ApplicationController < ActionController::Base
     redirect_to "/bo/pages/permission"
   end
 
+  def authenticate_and_authorize_active_user
+    return if skip_auth_on_this_controller?
+
+    authenticate_user!
+    redirect_to "/bo/pages/deactivated" if current_user_cannot_use_back_office?
+  end
+
   # http://jacopretorius.net/2014/01/force-page-to-reload-on-browser-back-in-rails.html
   def back_button_cache_buster
     response.headers["Cache-Control"] = "no-cache, no-store, max-age=0, must-revalidate"
     response.headers["Pragma"] = "no-cache"
     response.headers["Expires"] = "Fri, 01 Jan 1990 00:00:00 GMT"
+  end
+
+  private
+
+  def skip_auth_on_this_controller?
+    # Don't authorize and authenticate pages from HighVoltage or Devise
+    # Normally we'd use a skip_before_action, but these controllers are in gems
+    controller = params[:controller]
+    controller.include?("pages") || controller.include?("devise")
+  end
+
+  def current_user_cannot_use_back_office?
+    # Don't try to check user permissions if the user isn't logged in
+    return false if current_user.blank?
+
+    cannot? :use_back_office, :all
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,7 +40,7 @@ class ApplicationController < ActionController::Base
     return if skip_auth_on_this_controller?
 
     authenticate_user!
-    redirect_to "/bo/pages/deactivated" if current_user_is_deactivated?
+    redirect_to "/bo/pages/deactivated" if current_user_cannot_use_back_office?
   end
 
   # http://jacopretorius.net/2014/01/force-page-to-reload-on-browser-back-in-rails.html
@@ -59,7 +59,7 @@ class ApplicationController < ActionController::Base
     controller.include?("pages") || controller.include?("devise")
   end
 
-  def current_user_is_deactivated?
+  def current_user_cannot_use_back_office?
     # Don't try to check user permissions if the user isn't logged in
     return false if current_user.blank?
 
--- a/app/controllers/user_activations_controller.rb
+++ b/app/controllers/user_activations_controller.rb
@@ -3,36 +3,31 @@
 class UserActivationsController < ApplicationController
   include UsersHelper
 
+  before_action :assign_user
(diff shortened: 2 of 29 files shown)


irisfaraway:
We've got intermittent failures on this test in routes.rb:

```
  describe "GET /bo/[registration number]/renew" do
    context "when a deactivated user is signed in" do
      let(:user) { create(:user, :inactive) }
      let(:registration) { create(:registration, reg_identifier: "CBDU12345") }

      before(:each) do
        sign_in(user)
      end

      it "redirects to the deactivated page" do
        get "/bo/#{registration.reg_identifier}/renew"
        expect(response).to redirect_to("/bo/pages/deactivated")
      end
    end
  end
```

Sometimes it redirects to the permissions error page instead. Which is VERY ANNOYING. I think this is coming up because the FormsController in the engine has its own `authorize_user!` check and that sometimes gets called before the `before_action :authenticate_and_authorize_active_user` we have in the ApplicationsController here.

So we probably need to refactor how this is working in the engine, which will have an effect on the front office as well.

irisfaraway:
Did a rebase as quite a lot of changes have gone in since I last worked on this. Now can't seem to get the test failure any more. Yay?

app/controllers/user_activations_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def activate
+    assign_user(params[:id])
+    check_activation_permissions
+    @user.activate! unless @user.active?
+    redirect_to users_url
+  end
+
+  def deactivate
```
cintamani: I do believe that we shouldn't break the CRUD convention, and instead have 2 controllers, one for activation and one for deactivations

app/controllers/user_activations_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    assign_user(params[:id])
+    check_activation_permissions
```
cintamani: I do believe that stuff like `check_activation_permissions` should always be on top of the class as `before_action` rather than called in the methods

app/models/user.rb
```diff
@@ -7,6 +7,7 @@ class User
   # Use the User database
   store_in client: "users", collection: "back_office_users"
 
+  field :active, type: Boolean, default: true
```
cintamani: Since this is a new field, in order for all this to work we will need to either change the `active?` method to return `true` in case of `nil` or we can add a `deactivated` field to check instead, as Devise does

irisfaraway: Good spot, assumed that `default: true` would handle this but turns out it doesn't! Fixed.

app/controllers/application_controller.rb
```diff
@@ -35,10 +36,33 @@ def after_sign_out_path_for(*)
     redirect_to "/bo/pages/permission"
   end
 
+  def authenticate_and_authorize_active_user
+    return if skip_auth_on_this_controller?
+
+    authenticate_user!
+    redirect_to "/bo/pages/deactivated" if current_user_cannot_use_back_office?
```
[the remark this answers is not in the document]
irisfaraway: Renamed the method to clear things up 👍

app/controllers/user_activations_controller.rb
irisfaraway: Yep, can't remember why I have done this. Fixed!

app/controllers/user_activations_controller.rb
irisfaraway: I think this is a hangover from how it was done in WEX. But I agree. Will sort this out now.

irisfaraway: @cintamani All fixed!

app/controllers/application_controller.rb
```diff
@@ -35,10 +36,33 @@ def after_sign_out_path_for(*)
...
+  private
+
+  def skip_auth_on_this_controller?
+    # Don't authorize and authenticate pages from HighVoltage or Devise
+    # Normally we'd use a skip_before_action, but these controllers are in gems
+    controller = params[:controller]
+    controller.include?("pages") || controller.include?("devise")
+  end
+
+  def current_user_is_deactivated?
```
cintamani: I do find this method confusing. Can we change the name to: 
```
  def current_user_cannot_use_back_office?
```
Or, since this is used to redirect to a specific page that is `deactivated`, we can change the implementaation to be 
```
def current_user_is_deactivated?
  !(current_user && current_user.active?)
end
```
Actually by using the second approach, we can get rid of the extra code added in the Ability in regard to `can :use_back_office`

config/routes.rb
```diff
@@ -126,6 +126,24 @@
       to: "users#index",
       as: :users
 
+  resources :users,
+            only: [],
+            param: :id,
```
cintamani: Do we need to specify this param here? I think `:id` is the default

config/routes.rb
```diff
@@ -126,6 +126,24 @@
...
+            path: "/bo/users",
+            path_names: { show: "/:id" } do
+              resources :user_activations,
+                        as: :activations,
+                        only: %i[new create],
+                        path: "activate",
+                        path_names: { new: "" }
```
cintamani: Q: What do we need this `path_name` here for?

spec/requests/user_deactivations_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+
+        it "redirects to a permission error" do
+          get "/bo/users/#{active_user.id}/deactivate"
+          expect(response).to redirect_to("/bo/pages/permission")
+        end
+      end
+    end
+  end
+
+  describe "POST /bo/users/deactivate" do
```
cintamani:
```suggestion
  describe "POST /bo/users/:id/deactivate" do
```

spec/requests/user_deactivations_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+        sign_in(user)
+      end
+
+      context "when the current user has permission to activate the user" do
+        let(:current_user_role) { :finance_super }
+        let(:subject_user_role) { :finance }
+
+        context "when the user to be deactivated is active" do
+          it "redirects to the user list" do
+            post "/bo/users/#{active_user.id}/deactivate"
```
cintamani: Can we use url helpers with the tests? So if we change the path we won't need to fix all these as well :)

irisfaraway: I prefer having the URLs in there as then we can test that the path is as we expect too.

cintamani: Nice stuff :) But if we want to test the correctness of the path, then I suggest we go with creating specific route tests?

config/routes.rb
irisfaraway: I think it stops it tacking "new" on the end of the URL

spec/requests/user_deactivations_spec.rb
irisfaraway: Pretty much all the other request tests spell out the URL like this, so I'd rather stick with the convention for now I think.

config/routes.rb
```diff
@@ -126,6 +126,23 @@
       to: "users#index",
       as: :users
 
+  resources :users,
+            only: [],
+            path: "/bo/users",
+            path_names: { show: "/:id" } do
```
cintamani: Sorry just spotted this! Shall we get rid of this Path name as well? I am pretty sure is not used

config/routes.rb
cintamani: AAhhhh right! Yes I did have a vague record of why we used it but I wasn't sure :D We used it in the apps when we were dealing with the `identifiers` but we don't need it any more since tokens are a scoped route? I think we should keep it in this case, shall we?

irisfaraway:
IT'S THE INTERMITTENT FAILURE!
source

PR #521, opened 2019-12-18
Add main service for boxi export job

cintamani:

This includes:
 - New zip gem to create zip files
 - Implementation of the main service
 - Add rake task
- Add Boxi bucket configurations

Todo:
 - Implement serializers


.env.example
```diff
@@ -33,6 +33,7 @@ WCRS_EMAIL_SERVICE_EMAIL=""
 AWS_DAILY_EXPORT_ACCESS_KEY_ID=
 AWS_DAILY_EXPORT_SECRET_ACCESS_KEY=
 AWS_DAILY_EXPORT_BUCKET=
+AWS_BOXI_EXPORT_BUCKET=
```
Cruikshanks: Only comment is that these need to come in 3's

- Access Key ID
- Secret Access Key
- Boxi export bucket

We save a bit of time and money by using the same credentials and buckets in non-production environments, but often they are different when it comes to production.

So we need the flexibility to specify a different access key ID and secret access key for BOXI, as well as the bucket name.

cintamani: WIll add them in this PR :)

cintamani:
I am not sure what's actually wrong with Hakiri 🤔

irisfaraway:
@cintamani The Rack gem has a security vulnerability. I'll bump Dependabot and see if that updates it. Been patching other projects this morning.

irisfaraway:
Although actually looks like it should be fixed here: https://github.com/DEFRA/waste-carriers-back-office/pull/523

A rebase might fix it?

cintamani:
Just rebased! 🤞

PR #527, opened 2019-12-20
Add key people file creation on Boxi export

cintamani:

This adds the necessary code to generate the key_people.csv file for the boxi export

### DIFF
--- a/spec/models/reports/boxi/key_people_serializer_spec.rb
+++ b/spec/models/reports/boxi/key_people_serializer_spec.rb
@@ -28,7 +28,7 @@ module Reports
       describe "#add_entries_for" do
         let(:registration) { double(:registration) }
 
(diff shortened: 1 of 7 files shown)


spec/models/reports/boxi/key_people_serializer_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      before do
+        expect(CSV).to receive(:open).and_return(csv)
+        expect(csv).to receive(:<<).with(headers)
+      end
+
+      describe "#add_entries_for" do
+        let(:registration) { double(:registration) }
+
+        it "creates an entry in the csv for each address in the registration" do
```
Cruikshanks:
```suggestion
        it "creates an entry in the csv for each key person in the registration" do
```

PR #530, opened 2019-12-20
Add code to generate payments boxi export

cintamani:

this adds the code necessary to generate the payments.csv file included in the boxi export

### DIFF
--- a/spec/models/reports/boxi/payments_serializer_spec.rb
+++ b/spec/models/reports/boxi/payments_serializer_spec.rb
@@ -30,7 +30,7 @@ module Reports
       describe "#add_entries_for" do
         let(:registration) { double(:registration) }
 
-        it "creates an entry in the csv for each order_item in the registration" do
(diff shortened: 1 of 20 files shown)


spec/models/reports/boxi/payments_serializer_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+      before do
+        expect(CSV).to receive(:open).and_return(csv)
+        expect(csv).to receive(:<<).with(headers)
+      end
+
+      describe "#add_entries_for" do
+        let(:registration) { double(:registration) }
+
+        it "creates an entry in the csv for each order_item in the registration" do
```
Cruikshanks:
```suggestion
        it "creates an entry in the csv for each payment in the registration" do
```

cintamani: :joy:

PR #531, opened 2019-12-22
Implement generation of registration csv for Boxi

cintamani:

This adds the code that will generate the registrations.csv file for the boxi export.

### DIFF
--- a/spec/models/reports/boxi/registrations_serializer_spec.rb
+++ b/spec/models/reports/boxi/registrations_serializer_spec.rb
@@ -50,7 +50,7 @@ module Reports
       describe "#add_entries_for" do
         let(:registration) { double(:registration) }
 
-        it "creates an entry in the csv for each order_item in the registration" do
+        it "creates an entry in the csv for each registration" do
           presenter = double(:presenter)
 
           values = [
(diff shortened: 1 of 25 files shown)


spec/models/reports/boxi/registrations_serializer_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+
+      before do
+        expect(CSV).to receive(:open).and_return(csv)
+        expect(csv).to receive(:<<).with(headers)
+      end
+
+      describe "#add_entries_for" do
+        let(:registration) { double(:registration) }
+
+        it "creates an entry in the csv for each order_item in the registration" do
```
cintamani:
```suggestion
        it "creates an entry in the csv for each registration" do
```

spec/presenters/reports/boxi/registration_presenter_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+
+module Reports
+  module Boxi
+    RSpec.describe ::Reports::Boxi::RegistrationPresenter do
+      let(:registration) { double(:registration) }
+
+      subject { described_class.new(registration, nil) }
+
+      describe "#finance_details_balance" do
+        it "returns a formatted date" do
```
Cruikshanks: I'm thinking it does this 😁 !

```suggestion
        it "returns the balance formatted as pounds and pence" do
```

cintamani: Not really, I went back and fixed the usual one, but didn't spot this!

PR #532, opened 2019-12-22
Implement code for sign_offs csv Boxi file

cintamani:

Add code to generate the sign_offs.csv file for boxi export

### DIFF
--- a/spec/models/reports/boxi/sign_offs_serializer_spec.rb
+++ b/spec/models/reports/boxi/sign_offs_serializer_spec.rb
@@ -25,7 +25,7 @@ module Reports
       describe "#add_entries_for" do
         let(:registration) { double(:registration) }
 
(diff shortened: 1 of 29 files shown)


spec/models/reports/boxi/sign_offs_serializer_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+
+      before do
+        expect(CSV).to receive(:open).and_return(csv)
+        expect(csv).to receive(:<<).with(headers)
+      end
+
+      describe "#add_entries_for" do
+        let(:registration) { double(:registration) }
+
+        it "creates an entry in the csv for each order_item in the registration" do
```
cintamani:
```suggestion
        it "creates an entry in the csv for each conviction_sign_off in the registration" do
```

PR #554, opened 2020-01-03
Refunds - Implement service

cintamani:

This implements the service which will create the refund payment and the `new` view which trigger the post request to the service.

Screenshot 2020-01-03 at 18 05 28
Screenshot 2020-01-03 at 18 05 35
Screenshot 2020-01-03 at 18 05 43

### DIFF
--- a/spec/requests/refunds_spec.rb
+++ b/spec/requests/refunds_spec.rb
@@ -57,5 +57,35 @@ RSpec.describe "Refunds", type: :request do
   end
 
   # TODO
-  skip "POST /bo/finance_details/:_id/refunds/:order_key"
+  describe "POST /bo/finance_details/:_id/refunds/:order_key" do
+    context "when a valid user is signed in" do
+      let(:user) { create(:user) }
+      let(:renewing_registration) { create(:renewing_registration) }
+      let(:payment) { renewing_registration.finance_details.payments.first }
+
+      before(:each) do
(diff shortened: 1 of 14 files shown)


spec/requests/refunds_spec.rb
```diff
@@ -57,5 +57,35 @@
   end
 
   # TODO
```
irisfaraway:
```suggestion
```

spec/services/process_refund_service_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      expect(refund).to receive(:amount=).with(-1_500)
+      expect(refund).to receive(:registration_reference=).with("registration_reference")
+      expect(refund).to receive(:updated_by_user=).with("user@example.com")
+      expect(payment).to receive(:worldpay?).and_return(worldpay)
+    end
+
+    context "when the payment is a card payment" do
+      let(:worldpay) { true }
+
+      it "generates a new refund payment and associate it with the right finance details" do
```
irisfaraway: Maybe this is my foggy post-holiday brain, but where are we testing that the payment is successfully generated and assigned to the finance details? Same question for the cash context.

cintamani: Yes we are :) is just all mocks and stubbing since this is an integration test. Most of the test code is in the `before` context

PR #559, opened 2020-01-07
Add permissions checks to refund

cintamani:

This sorts out permissions checks on the refund functionality

### DIFF
--- a/spec/support/shared_examples/abilities/agency_with_refund_examples.rb
+++ b/spec/support/shared_examples/abilities/agency_with_refund_examples.rb
@@ -4,4 +4,8 @@ RSpec.shared_examples "agency_with_refund examples" do
   it "should be able to view revoked reasons" do
     should be_able_to(:view_revoked_reasons, WasteCarriersEngine::RenewingRegistration)
   end
+
+  it "should be able to refund a payment" do
+    should be_able_to(:refund, WasteCarriersEngine::Registration)
+  end
 end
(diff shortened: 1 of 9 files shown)


spec/support/shared_examples/abilities/agency_with_refund_examples.rb
```diff
@@ -4,4 +4,8 @@
   it "should be able to view revoked reasons" do
     should be_able_to(:view_revoked_reasons, WasteCarriersEngine::RenewingRegistration)
   end
+
+  it "should be able to refund a payment" do
```
irisfaraway: Can we also test that the finance users and the "below agency_with_refund examples" cannot do this?

cintamani: Oh yes! Good spot :)

spec/support/shared_examples/abilities/below_agency_with_refund_examples.rb
```diff
@@ -4,4 +4,8 @@
   it "should not be able to view revoked reasons" do
     should_not be_able_to(:view_revoked_reasons, WasteCarriersEngine::RenewingRegistration)
   end
+
+  it "should not be able to refund a payment" do
```
irisfaraway: This example doesn't cover the finance user types.

cintamani: I am not sure I get the structure of those tests yet 😅 Added more

PR #561, opened 2020-01-08
Calculate the correct amount for refund

cintamani:

We have discovered that the refund should only trigger for outstanding amounts in registrations. This fixes the refund services so that the correct amount is refunded. In case the payment selected for the refund has an amount that is less than the overpaid balance, the refund amount will be set to the payment amount.

### DIFF
--- a/app/services/process_refund_service.rb
+++ b/app/services/process_refund_service.rb
@@ -2,9 +2,11 @@
 
 class ProcessRefundService < WasteCarriersEngine::BaseService
   def run(finance_details:, payment:, user:)
+    @finance_details = finance_details
     @payment = payment
     @user = user
 
+    return false if amount_to_refund == 0
     return false if card_payment? && !card_refund_result
 
     finance_details.payments << build_refund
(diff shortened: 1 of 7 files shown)


irisfaraway:
@cintamani Looks like there is a Rubocop issue! Can you fix that and then I'll have a look at it? 🙏

app/services/process_refund_service.rb
```diff
@@ -2,9 +2,11 @@
 
 class ProcessRefundService < WasteCarriersEngine::BaseService
   def run(finance_details:, payment:, user:)
+    @finance_details = finance_details
     @payment = payment
     @user = user
 
+    return false if amount_to_refund == 0
```
cintamani:
```suggestion
    return false if amount_to_refund.zero?
```

spec/services/process_refund_service_spec.rb
```diff
@@ -4,22 +4,31 @@
...
     let(:user) { double(:user, email: "user@example.com") }
     let(:payments) { double(:payments) }
     let(:refund) { double(:refund) }
+    let(:worldpay) { true }
 
     before do
       allow(payment).to receive(:worldpay?).and_return(worldpay)
     end
 
+    context "when the amount to refund is 0" do
```
Cruikshanks: I think I'd benefit from tweaking the context title. We _know_ the amount to refund is 0 only because we know logic inside the service will calculate it like that because we have a positive balance.

So to me what we are actually testing is that if the balance is not negative it returns false and does not create a payment.

```suggestion
    context "when the registration balance is not negative" do
```

cintamani:
Sorry guys! looks like something is broken in the test!

app/services/process_refund_service.rb
```diff
@@ -17,17 +19,26 @@ def run(finance_details:, payment:, user:)
...
   def card_refund_result
-    @_card_refund_result ||= ::Worldpay::RefundService.run(payment: payment)
+    @_card_refund_result ||= ::Worldpay::RefundService.run(payment: payment, amount: amount_to_refund)
+  end
+
+  def amount_to_refund
+    # We can never refund unless there have been an overpayment.
+    return 0 unless finance_details.balance.negative?
+
+    overpayment = (-1 * finance_details.balance)
```
Cruikshanks: Purely because I'm slow on the uptake could we add a comment that just explains what this line is doing?

```suggestion
    # A quick maths trick to convert a negative value to a postive one
    overpayment = (-1 * finance_details.balance)
```

PR #563, opened 2020-01-08
Fix code to render error flash messages

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-821

I have just realised that I have never implemented the actual code to show the error flash messages. This fixes it as now we will have case of errors since we are dealing with requests to third party API systems

Screenshot 2020-01-08 at 22 23 42

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -56,7 +56,7 @@ gem "defra_ruby_aws", "~> 0.2.0"
 # Use the waste carriers engine for the user journey
 gem "waste_carriers_engine",
     git: "https://github.com/DEFRA/waste-carriers-engine",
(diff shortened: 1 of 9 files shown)


Gemfile
```diff
@@ -56,7 +56,7 @@ gem "defra_ruby_aws", "~> 0.2.0"
 # Use the waste carriers engine for the user journey
 gem "waste_carriers_engine",
     git: "https://github.com/DEFRA/waste-carriers-engine",
-    branch: "master"
+    branch: "add-error-flash-partial"
```
cintamani:
```suggestion
    branch: "master"
```

PR #569, opened 2020-01-09
Change back office user role

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-761

It should be possible for super users to change the roles of other back office users.

This is dependent on the user group - for example, an agency super can only change the role of other agency users. The roles which are available to choose are also limited by the super user's group - so a finance super can only select finance roles.


irisfaraway:
Running into failures on the request test because the password isn't validating properly. We got around this with `set` and `update_attribute` previously, but I'd like to see if there is a better solution that still calls our validations.

app/controllers/user_activity_controller.rb
```diff
@@ -4,7 +4,7 @@ class UserActivityController < ApplicationController
   include UsersHelper
 
   before_action :assign_user
-  before_action :check_activation_permissions
+  before_action :check_modification_permissions
```
cintamani: nice one :)

app/models/concerns/can_behave_like_user.rb
```diff
@@ -47,7 +47,6 @@ module CanBehaveLikeUser
     field :locked_at,       type: Time
 
     # Validations
-    validates :password, presence: true, length: { in: 8..128 }
```
cintamani: Q: Why taking this away?

app/models/user.rb
```diff
@@ -37,21 +37,28 @@ def deactivate!
...
-             finance_super].freeze
+  AGENCY_ROLES = %w[agency
+                    agency_with_refund
+                    agency_super].freeze
+
+  FINANCE_ROLES = %w[finance
+                     finance_admin
+                     finance_super].freeze
+
+  ROLES = (AGENCY_ROLES + FINANCE_ROLES).freeze
```
cintamani: 💯

app/views/shared/_select_role.html.erb
```diff
@@ -0,0 +1,22 @@
+<% if resource.errors[:role].any? %>
+<div class="form-group form-group-error">
+<% else %>
+<div class="form-group">
+<% end %>
```
cintamani:
```suggestion
<div class="form-group <%= "form-group-error" if resource.errors[:role].any? %>">
```

app/views/shared/_select_role.html.erb
```diff
@@ -0,0 +1,22 @@
...
+  <fieldset id="role">
+    <legend>
+      <%= t(".roles.legend") %>
+    </legend>
+
+    <% if resource.errors[:role].any? %>
+    <span class="error-message"><%= resource.errors[:role].join(", ") %></span>
```
cintamani:
```suggestion
      <span class="error-message"><%= resource.errors[:role].join(", ") %></span>
```

config/locales/en.yml
```diff
@@ -34,6 +34,17 @@ en:
       name: "Name"
       position: "Position"
       dob: "Date of birth"
+      shared:
```
cintamani: To delete? :) 
```suggestion
```

irisfaraway: Whoops! Rebase error 🤦‍♀

app/models/concerns/can_behave_like_user.rb
irisfaraway: Did you see my comment on https://github.com/DEFRA/waste-carriers-back-office/pull/569/commits/c0244cfe47869033510724b066ac4bc01b63c712?

PR #570, opened 2020-01-10
Edit permissions for cease and revoke

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-807

Chris has provided further clarification on who should be able to cease/revoke:

You are right to ask.  It should be anyone in the Carriers admin team so agency with refund users


spec/support/shared_examples/abilities/agency_super_examples.rb
```diff
@@ -8,4 +8,12 @@
   it "should be able to manage agency users" do
     should be_able_to(:manage_agency_users, User)
   end
+
```
irisfaraway: I don't think you need these ones in the `super` examples as well, since the ability spec runs the agency_with_refund tests for agency supers too.

PR #572, opened 2020-01-10
[WIP] Do not show link to finance page if there is no finance details

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-808

QA have reported an error thrown by the page being visited from a transient registration that was still in progress. This was due to the fact that the page expect the presence of a `finance_details` object in the registration, but in case of in-progress registration there is none. Hence we shouldn't show the link there.

### DIFF
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -58,6 +58,7 @@ module ActionLinksHelper
   end
 
   def display_refund_link_for?(resource)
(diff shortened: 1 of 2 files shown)


cintamani:
Going to add a new link helper for details link, as currently this helper is shared between the process payment and finance details

cintamani:
Closing this as the change have been made in https://github.com/DEFRA/waste-carriers-back-office/pull/576

PR #574, opened 2020-01-10
Add 'invite new user' feature

irisfaraway:

Back office super users should be able to invite new back office users.

Users will be expected to enter the email address for the new user, select a role, and when submitted behind the scenes the app will create a new user and send an email with a link to the email. The link validates the email address and presents a screen that asks the user to create a password.

Super users should only be able to invite users with roles of their user group - for example, agency supers can invite new agency users, but not new finance users.

### DIFF
--- a/app/controllers/user_invitations_controller.rb
+++ b/app/controllers/user_invitations_controller.rb
@@ -14,4 +14,8 @@ class UserInvitationsController < Devise::InvitationsController
   def configure_permitted_parameters
     devise_parameter_sanitizer.permit(:invite, keys: [:role])
   end
+
+  def after_invite_path_for(_resource)
+    users_path
+  end
 end
--- a/app/controllers/user_invitations_controller.rb
+++ b/app/controllers/user_invitations_controller.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class UserInvitationsController < Devise::InvitationsController
+  before_action :configure_permitted_parameters
+
+  private
+
+  # This allows us to include a role on the user invitation form
+  def configure_permitted_parameters
+    devise_parameter_sanitizer.permit(:invite, keys: [:role])
+  end
+end
--- a/app/controllers/user_invitations_controller.rb
+++ b/app/controllers/user_invitations_controller.rb
@@ -1,9 +1,14 @@
 # frozen_string_literal: true
 
 class UserInvitationsController < Devise::InvitationsController
+  before_action :authorize, only: %i[new create]
   before_action :configure_permitted_parameters
 
   private
+
+  def authorize
+    authorize! :manage_back_office_users, current_user
+  end
 
   # This allows us to include a role on the user invitation form
   def configure_permitted_parameters
--- a/app/controllers/user_invitations_controller.rb
+++ b/app/controllers/user_invitations_controller.rb
@@ -7,7 +7,7 @@ class UserInvitationsController < Devise::InvitationsController
   before_action :configure_permitted_parameters
 
   def create
-    if selected_role_is_valid?(params.dig(:user, :role))
+    if selected_role_is_in_allowed_group?(params.dig(:user, :role))
       super
     else
       new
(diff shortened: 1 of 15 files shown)


irisfaraway:
Will need a rebase after https://github.com/DEFRA/waste-carriers-back-office/pull/569 is merged, but otherwise ready to review.

app/controllers/user_invitations_controller.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class UserInvitationsController < Devise::InvitationsController
+  include UsersHelper
+
+  before_action :authorize, only: %i[new create]
+  before_action :configure_permitted_parameters
+
+  def create
+    if selected_role_is_in_allowed_group?(params.dig(:user, :role))
```
cintamani: `#dig` 😍

app/views/devise/invitations/edit.html.erb
```diff
@@ -0,0 +1,45 @@
...
+         <%= t(".heading") %>
+       </h1>
+
+       <p><%= t(".paragraph_1") %></p>
+
+      <%= render("waste_carriers_engine/shared/errors", object: resource) %>
+
+      <%= f.hidden_field :invitation_token %>
+
+      <% if f.object.class.require_password_on_accepting %>
```
cintamani: Do we need to check this in our system? I don't think we will have any instance where this is going to be false?

app/views/users/index.html.erb
```diff
@@ -36,10 +37,12 @@
                   <%= t(".user_list.roles.#{user.role}") %>
                 </td>
                 <td>
-                  <% if user.active? %>
-                    <%= t(".user_list.statuses.active") %>
-                  <% else %>
+                  <% if !user.active? %>
```
cintamani: I think we need a `User#deactivated?`

config/environments/development.rb
```diff
@@ -53,7 +53,7 @@
   # Raises error for missing translations
   config.action_view.raise_on_missing_translations = true
 
-  config.action_mailer.default_url_options = { host: "localhost", port: 3000 }
+  config.action_mailer.default_url_options = { host: config.wcrs_back_office_url, protocol: "http" }
```
cintamani: Do we need these settings to be merged? Are they part of the work here?

config/environments/test.rb
```diff
@@ -32,6 +32,8 @@
   # The :test delivery method accumulates sent emails in the
   # ActionMailer::Base.deliveries array.
   config.action_mailer.delivery_method = :test
+  # Required for testing user invitations
+  config.action_mailer.default_url_options = { host: config.wcrs_back_office_url, protocol: "http" }
```
cintamani: Same as above :)

config/environments/development.rb
irisfaraway: Yep, see comment on https://github.com/DEFRA/waste-carriers-back-office/pull/574/commits/f5f3831d5f996e7b3c6bf45a34fac6a0a9b7f7d7! Otherwise this didn't work locally.

config/environments/test.rb
irisfaraway: Yes, tests fail without this setting. This follows what we had to do for WEX as well.

app/views/devise/invitations/edit.html.erb
irisfaraway: Good point, will remove that.

app/views/users/index.html.erb
irisfaraway: Yeah, that would be nicer to read! Might do this in a separate PR to stop this one getting too bulky :)

app/views/devise/invitations/edit.html.erb
irisfaraway: Fixed! https://github.com/DEFRA/waste-carriers-back-office/pull/574/commits/8eb3374a1b2282b071ace36afde84a5c0d331aae

PR #576, opened 2020-01-13
Add new method for finance_details check

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-808

This fixes a situation where we were using the same helper method for both the `add payment` and `finance details` link show. This also updates the tests to use `resource` rather than `resource` variable names.

### DIFF
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -63,12 +63,8 @@ module ActionLinksHelper
     can?(:refund, resource)
   end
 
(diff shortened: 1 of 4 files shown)


app/helpers/action_links_helper.rb
```diff
@@ -63,12 +63,8 @@ def display_refund_link_for?(resource)
     can?(:refund, resource)
   end
 
-  def display_payment_details_link_for?(resource)
```
cintamani: This was never used

PR #588, opened 2020-01-16
Add missing titles to new user invite pages

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-683

The 'Invite a new user' and 'Create a new back office account' pages from the process to add a new back office user did not have title attributes. This PR adds them.

### DIFF
--- a/app/views/devise/invitations/new.html.erb
+++ b/app/views/devise/invitations/new.html.erb
@@ -1,3 +1,5 @@
+<% content_for :title, t(".title")  %>
+
 <div class="grid-row">
   <div class="column-full">
     <%= render("waste_carriers_engine/shared/back", back_path: users_path) %>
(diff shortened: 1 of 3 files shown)


app/views/devise/invitations/new.html.erb
```diff
@@ -1,3 +1,5 @@
+<% content_for :title, t(".title")  %>
```
cintamani: I think that the code in the application layout already add a title if present in the translation model without the need of this code

irisfaraway: Yeah, I tried that first but it didn't seem to work with the Devise bits. I'll stop being lazy and try a bit harder to make it work 😛

cintamani: Oh! That's odd! If it is going to get a painful bit, don't worry! I am happy to approve this if we decide that is too much hassle :D

irisfaraway: @cintamani Worked it out - I had forgotten that since we had our own UserInvitationsController it would need a different name in the YAML.

PR #590, opened 2020-01-16
Fix default email protocol

irisfaraway:

During QA of the 'invite user' feature, we spotted that the email confirmation link did not work as expected when opened directly into incognito mode.

This turned out to be because of the default protocol being `http`. This may just be an issue on our test environments, but we should default to `https` anyway.

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -16,7 +16,7 @@ Rails.application.configure do
   config.action_controller.perform_caching = false
 
   # Sending e-mails is required for user management and registration e-mails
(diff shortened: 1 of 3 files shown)


Cruikshanks:
@irisfaraway Do you know what the impact is when working locally? Will it still work there and in the vagrant box?

irisfaraway:
@Cruikshanks The default localhost URLs in the application.rb config include `http`, so it still uses those in the URLs when I run it on Vagrant.

PR #595, opened 2020-01-19
Finance - Write Offs

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-810

This PR implements the write off feature on the back office. 

Each commit is self-contained, I suggest a review commit-by-commit. 

Screenshot 2020-01-19 at 13 59 09
Screenshot 2020-01-19 at 13 59 16
Screenshot 2020-01-19 at 13 59 31
Screenshot 2020-01-19 at 14 11 14


Gemfile
```diff
@@ -56,7 +56,7 @@ gem "defra_ruby_aws", "~> 0.2.0"
 # Use the waste carriers engine for the user journey
 gem "waste_carriers_engine",
     git: "https://github.com/DEFRA/waste-carriers-engine",
-    branch: "master"
+    branch: "balance-difference-from-zero"
```
cintamani:
```suggestion
    branch: "master"
```

spec/services/process_write_off_service_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+        unpaid_balance: unpaid_balance,
+        payments: payments
+      )
+    end
+    let(:comment) { double(:comment) }
+    let(:user) { double(:user, email: double(:email)) }
+
+    let(:result) { described_class.run(finance_details: finance_details, user: user, comment: comment) }
+
+    before do
```
irisfaraway: Can we delete this empty `before` block?

spec/support/shared_examples/abilities/agency_examples.rb
```diff
@@ -16,6 +16,34 @@
     should be_able_to(:review_convictions, WasteCarriersEngine::RenewingRegistration)
   end
 
+  context ":write_off_small" do
```
irisfaraway: I think we may be missing some tests for abilities and user types – agency users only have tests for `:write_off_small`, but not `:write_off_large`. Meanwhile some finance users have tests for `:write_off_large` (although I think base finance users have none) and none of them have tests for `:write_off_small`.

spec/support/shared_examples/abilities/agency_examples.rb
```diff
@@ -16,6 +16,34 @@
...
+    let(:finance_details) { build(:finance_details, balance: balance) }
+
+    context "when the zero difference balance is less than the setted cap" do
```
irisfaraway:
```suggestion
    context "when the zero-difference balance is less than the write-off cap for agency users" do
```

[NIT] Maybe personal preference, but I think this is clearer about what the cap is. The rewording would apply to some other specs as well.

app/services/process_write_off_service.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  private
+
+  attr_reader :user, :finance_details
+
+  def build_write_off(comment)
+    write_off = WasteCarriersEngine::Payment.new(payment_type: payment_type)
+
+    write_off.order_key = SecureRandom.uuid.split("-").last
```
irisfaraway: What do we do with this `order_key`?

app/forms/write_off_form.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    return false unless valid?
+
+    ProcessWriteOffService.run(
+      finance_details: transient_registration.finance_details,
+      user: user,
+      comment: comment
+    )
+
+    true
```
irisfaraway: Should we drop this `true` and return whatever the `ProcessWriteOffService` does? Currently the service returns true or false depending on the success but we don't seem to do anything with that.

spec/services/process_write_off_service_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+
+          expect(payments).to receive(:<<).with(write_off)
+          expect(finance_details).to receive(:update_balance)
+          expect(finance_details).to receive(:save!)
+
+          result
+        end
+      end
+
+      context "when the user has not write_off_large permissions" do
```
irisfaraway:
```suggestion
      context "when the user does not have write_off_large permissions" do
```

[NIT] Just a typo :)

app/controllers/write_offs_controller.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+class WriteOffsController < ApplicationController
```
irisfaraway: I think we'd talked about using AdminFormsController for this last week – did it not work out? Or are we planning to do some separate refactoring?

I would be tempted to rename this to `WriteOffFormsController` as I think all the other controllers using form objects include `Form` in the name. So it might make it more consistent and be a good indicator that it uses a form object.

spec/support/shared_examples/abilities/agency_examples.rb
cintamani: Finance admin and finance super can do write off large, only finance-user cannot. Hence I think that's correct :) Will add more anyway! I did wrote these tests before I decided it was a good idea to split them off as suggested by Alan :D

app/services/process_write_off_service.rb
cintamani: Nothing at all. But the old system generates it like that when doing write off, so for now, I have added it as per old system

spec/support/shared_examples/abilities/agency_examples.rb
irisfaraway: The users that can do things have tests that they can, but there are no tests that users are not able to do things they don't have permission to do. If that makes sense!

app/forms/write_off_form.rb
cintamani: I have thought about that, but it generates a dependency between this method/class and the service itself, which will impact negatively to maintainability

app/controllers/write_offs_controller.rb
cintamani: We did talk about the AdminFormsController in regard to transient payments, no refunds. I am not convinced about using it here yet. Currently, here I don't use anything of the things implemented in AdminFormsController. There is still a lot of code there that is dependant on renewals, and I need to refactor that out first.

irisfaraway: Ahhh, that was it – I was thinking of the wrong feature 🤦‍♀ Ignore me then!

app/forms/write_off_form.rb
irisfaraway: I see where you are coming from. However it does mean that right now if the balance is 0, the service returns `false` and we don't do anything, but as far as the user knows they have successfully made a change. So I worry that could cause some confusion.

cintamani: I don't really worry about it, as the button for doing a refund won't show if the balance is 0 they can only trigger the feature hacking the URL somehow :D

cintamani: We do return `false` but is not meaningful for now, is there in case of future development on the feature :)

irisfaraway: OK, if they shouldn't be accessing it anyway that's alright with me for now.

spec/requests/write_off_forms_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 require "rails_helper"
 
-RSpec.describe "WriteOffs", type: :request do
+RSpec.describe "WriteOffForms", type: :request do
   describe "GET /bo/finance_details/:_id/write_off/new" do
```
irisfaraway: Is this URL still correct?

config/locales/write_offs.en.yml
```diff
@@ -1,5 +1,5 @@
 en:
```
irisfaraway: Can we rename this file as well? 🙏

spec/support/shared_examples/abilities/agency_examples.rb
```diff
@@ -16,6 +16,38 @@
...
+
+    context "when the zero-difference balance is less than the write-off cap for agency users" do
+      let(:balance) { WasteCarriersBackOffice::Application.config.write_off_agency_user_cap.to_i - 1 }
+
+      it "should be able to write off" do
+        should be_able_to(:write_off_small, finance_details)
+      end
+    end
+
+    context "when the zero difference balance is more than the setted cap" do
```
irisfaraway:
```suggestion
    context "when the zero-difference balance is more than the write-off cap for agency users" do
```

spec/support/shared_examples/abilities/agency_examples.rb
```diff
@@ -16,6 +16,38 @@
...
+      let(:balance) { WasteCarriersBackOffice::Application.config.write_off_agency_user_cap.to_i + 1 }
+
+      it "should not be able to write off" do
+        should_not be_able_to(:write_off_small, finance_details)
+      end
+    end
+
+    context "when the zero difference balance is equal to the setted cap" do
```
irisfaraway:
```suggestion
    context "when the zero-difference balance is equal to the write-off cap for agency users" do
```

config/locales/write_offs.en.yml
```diff
@@ -0,0 +1,18 @@
+en:
```
irisfaraway: Can we rename this file to `write_off_forms.en.yml`?

spec/requests/write_off_forms_spec.rb
cintamani: Good catch!! I will change the route path though as I think is better for the user to read this than `/bo/finance_details/:_id/write_off_form/new`

PR #600, opened 2020-01-20
Add resource scope

cintamani:
During one of our meetings, we have decided that it is time now for us to have a `resource` scope which is able to take an `id` param and use it to search in multiple registration collections. 
This refactors the: `:finance_details`, `:refunds` and `:write_offs` resource under the new `resources`. 

Suggest going commit-by-commit for review.


spec/requests/write_off_forms_spec.rb
```diff
@@ -3,7 +3,7 @@
 require "rails_helper"
 
 RSpec.describe "WriteOffForms", type: :request do
-  describe "GET /bo/finance_details/:_id/write_off/new" do
+  describe "GET /bo/resource/:_id/write_off/new" do
```
irisfaraway:
```suggestion
  describe "GET /bo/resource/:_id/write-off/new" do
```

spec/requests/write_off_forms_spec.rb
```diff
@@ -57,14 +57,14 @@
...
-        get new_finance_details_write_off_form_path("foo")
+        get new_resource_write_off_form_path("foo")
 
         expect(response).to redirect_to(new_user_session_path)
       end
     end
   end
 
-  describe "POST /bo/finance_details/:_id/write_off" do
+  describe "POST /bo/resource/:_id/write_off" do
```
irisfaraway:
```suggestion
  describe "POST /bo/resource/:_id/write-off" do
```

PR #601, opened 2020-01-20
Add page to import convictions

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-772

We want to be able to update conviction check data through the app, rather than using the existing Java service or directly updating the database.

For this we want a new page with a text box, some content that explains what it's for, a warning and a continue button. Submitting a valid CSV in this text box should call the ConvictionImportService and update the data.

Only developer users should be able to access this feature.


irisfaraway:
I don't know what CodeClimate wants from me as everything new has 100% test coverage 😑 but this is ready for humans to review!

PR #604, opened 2020-01-21
Update mocks engine to support Worldpay

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-818

The [defra-ruby-mocks](https://github.com/DEFRA/defra-ruby-mocks) engine that we are using to support mocking external services now supports Worldpay.

This change updates the engine and adds the additional config to allow the back-office to stand in as a mock for Worldpay.

### DIFF
--- a/config/initializers/defra_ruby_mocks.rb
+++ b/config/initializers/defra_ruby_mocks.rb
@@ -7,4 +7,15 @@ DefraRubyMocks.configure do |configuration|
   # Set how long the mock should delay before responding. In the engine itself
   # the default is 1000ms (1 second)
   configuration.delay = ENV["WCRS_MOCK_DELAY"] || 1000
+
+  # Tell the mocks engine details needed to mock worldpay. These are needed
+  # so it can then generate values that the calling app will verify as valid
+  configuration.worldpay_admin_code = ENV["WCRS_WORLDPAY_ADMIN_CODE"]
+  configuration.worldpay_merchant_code = ENV["WCRS_WORLDPAY_ECOM_MERCHANTCODE"]
+  configuration.worldpay_mac_secret = ENV["WCRS_WORLDPAY_ECOM_MACSECRET"]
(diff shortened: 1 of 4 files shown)


config/initializers/defra_ruby_mocks.rb
```diff
@@ -7,4 +7,15 @@
   # Set how long the mock should delay before responding. In the engine itself
   # the default is 1000ms (1 second)
   configuration.delay = ENV["WCRS_MOCK_DELAY"] || 1000
+
+  # Tell the mocks engine details needed to mock worldpay. These are needed
+  # so it can then generate values that the calling app will verify as valid
+  configuration.worldpay_admin_code = ENV["WCRS_WORLDPAY_ADMIN_CODE"]
+  configuration.worldpay_merchant_code = ENV["WCRS_WORLDPAY_ECOM_MERCHANTCODE"]
```
irisfaraway: Should these be MOTO for the back office?

Cruikshanks: It doesn't really matter! 😁 

Basically, because I had to go forth with the front and back-office apps talking to each other when it comes to the mocking, the important thing is that they use the same values.

So to start with, knowing the back-office is the mock for the front-office I chose to use the ECOM values as that is what the front-office uses.

After testing with the frontend app though, I realised that I have one app talking to one mock Worldpay endpoint (the front-office), however, it's expecting a response generated using both the ECOM and MOTO credentials depending on whether it's an external or internal user.

😦 💥 

The only solution was when mocking to ensure that the MOTO and ECOM merchant and mac codes matched. So, in the end, it doesn't really matter what env vars I use here. So I opted to go for a world where the frontend doesn't exist!

I hope that makes sense. Happy to discuss if not.

irisfaraway: @Cruikshanks That sounds sensible to me. Thanks!

PR #614, opened 2020-01-22
Add guard clause to missing object

cintamani:

This adds guard clause against broken data to boxi export


cintamani:
Codeclimate is a won't fix

PR #625, opened 2020-01-29
More fixes to BOXI

cintamani:
In order for our file to match the one generated by the old system, we haad to make a couple of fixes: 
- Sanitize data and replace `\n` with `.`
- Start counting our UID from 1 rather than from 0


cintamani:
Codeclimate is a won't fix

PR #628, opened 2020-01-29
Fix changelog generator

Cruikshanks:
It seems since [github-changelog-generator](https://github.com/github-changelog-generator/github-changelog-generator) v1.15.0 dropped last year that the generator has lost the ability to determine the `user` and `project` from the current repo its run in.

Now when we run it we have to call the gem directly and pass these in as arguments to get it to work.

So this change resolves the issue, and means we can go back to using `bundle exec rake changelog`.


irisfaraway:
Think these test failures may be a daylight savings issue from running after 23:00

PR #630, opened 2020-01-30
Actually fix redirect after accepting user invite

irisfaraway:

Previously BO users who created their an account through an invitation would be redirected to the old backend after setting a password. We want to redirect them to the new back office dashboard instead.

My last fix (https://github.com/DEFRA/waste-carriers-back-office/pull/617) didn't work as expected, so hoping this one will actually do the trick. This time we have tests!


irisfaraway:
Tested this out on dev and it woooooooorks 🎉

PR #640, opened 2020-02-03
Boxi - Unique UID for orders

cintamani:

We have realised that order's UID are not per-registration but instead global, faking a proper SQL autoincrementing integer ID.
In order for us to reproduce that same logic, we have decided to twick the main BOXI service so that it has a knowledge of orders UID as well as Registrations UID. Those have now to be passed to the serializers as a separate attribute, and serializers on orders are now dealing with one order at a time only.


spec/models/reports/boxi/order_items_serializer_spec.rb
```diff
@@ -55,17 +51,14 @@ module Boxi
...
           expect(csv).to receive(:<<).with(values)
 
-          subject.add_entries_for(registration, 0)
+          subject.add_entries_for(order, 0, 0)
         end
 
         it "sanitize data before inserting them in the csv" do
-          order = double(:order)
-          presenter = double(:presenter, description: " string to\r\nsanitize\n").as_null_object
+          presenter = double(:presenter, description: "string to sanitize\n").as_null_object
```
cintamani:
```suggestion
          presenter = double(:presenter, description: " string to\r\nsanitize\n").as_null_object
```

PR #642, opened 2020-02-03
View registration certificate in back office

irisfaraway:

We already email out certificates from the back office with completed renewals.

However, the old backend also has a feature where you can view a certificate on-demand through the back office. The PDF generation is already in place in the engine. We just need to load it from a route.


irisfaraway:
Hakiri build passed, so going ahead and merging this https://hakiri.io/github/DEFRA/waste-carriers-back-office/master/e8f71c1a3fd9098b533e02814681069f6c5127a2

PR #644, opened 2020-02-04
Finance - Charge adjust

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-870

This PR contains the code for positive and negative charge adjust. 
Some explanations on the choices I have made: 
 - I have used forms which currently inherits from the Engine base form, even though I only need basic setups and I do nothing with the resource object itself. In the future, if we generate resources forms, we can decide to inherit from a simpler model.
 - I have decided to not reuse exactly the logic to validate and cast a string to integer used for the other payments form. This is because in the case when a user insert invalid characters that system currently fails with a 500 error while the old system has a custom validation of `Invalid format`, hence with Andrew we agreed we would validate the format as per old system and keep the field as text. 
- There is a `start_form` for the charge adjust, but no state machine. I have decided that would be an overkill, as I don't really need to persist the `positive` VS `negative` choice hence I don't need a transient object. However, it is arguable that the controller looks a bit hacky as it does not use the standard Resource form submit implementation. Happy to discuss / change approach to it :) 

*Reviewers:* I have made this a commit-by-commit reviewable PR :D


app/controllers/charge_adjust_start_forms_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+  def submit_form(form, params)
+    if form.submit(params)
+      redirect_to public_send("new_resource_#{@charge_adjust_start_form.charge_type}_charge_adjust_form_path")
+    else
+      render :new
+    end
+  end
+
+  def authorize_user
+    raise CanCan::AccessDenied unless can?(:charge_adjust, @resource)
```
cintamani: Use autorize!

app/controllers/positive_charge_adjust_forms_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  private
+
+  def positive_charge_adjust_form_params
+    params.fetch(:positive_charge_adjust_form, {}).permit(:amount, :reference, :description)
+  end
+
+  def authorize_user
+    raise CanCan::AccessDenied unless can?(:charge_adjust, @resource)
```
cintamani: use autorize!

app/controllers/negative_charge_adjust_forms_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  private
+
+  def negative_charge_adjust_form_params
+    params.fetch(:negative_charge_adjust_form, {}).permit(:amount, :reference, :description)
+  end
+
+  def authorize_user
+    raise CanCan::AccessDenied unless can?(:charge_adjust, @resource)
```
cintamani: use autorize!

app/forms/concerns/can_convert_string_to_cents.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+module CanConvertStringToCents
```
irisfaraway:
```suggestion
module CanConvertStringToPence
```

app/services/process_charge_adjust_service.rb
```diff
@@ -0,0 +1,36 @@
...
+    finance_details.update_balance
+    finance_details.save!
+  end
+
+  private
+
+  attr_reader :form, :user
+
+  def build_order
+    order_item = WasteCarriersEngine::OrderItem.new_charge_adjust_item
```
cintamani: Create a method for creating `order_item`

PR #646, opened 2020-02-04
Display 'view certificate' actions link

irisfaraway:

Now that we have the certificate in the back office, we can display this action link again.

It also renames the methods to be more accurate.

### DIFF
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -41,6 +41,12 @@ module ActionLinksHelper
     new_registration_registration_transfer_path(resource.reg_identifier)
   end
 
+  def certificate_link_for(resource)
+    return "#" unless a_registration?(resource)
+
+    registration_certificate_path(resource.reg_identifier)
+  end
+
   def display_details_link_for?(resource)
     a_transient_registration?(resource) || a_registration?(resource)
   end
@@ -99,9 +105,7 @@ module ActionLinksHelper
     can?(:update, WasteCarriersEngine::Registration)
   end
 
-  def display_view_confirmation_letter_link_for?(resource)
-    # TODO: delete next line filter when internal route exists https://eaflood.atlassian.net/browse/RUBY-786
-    return false if a_registration?(resource)
+  def display_certificate_link_for?(resource)
     return false unless display_registration_links?(resource)
     return false unless can?(:view_certificate, WasteCarriersEngine::Registration)
 
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -41,12 +41,6 @@ module ActionLinksHelper
     new_registration_registration_transfer_path(resource.reg_identifier)
   end
 
-  def certificate_link_for(resource)
-    return "#" unless a_registration?(resource)
-
-    registration_certificate_path(resource.reg_identifier)
-  end
-
   def display_details_link_for?(resource)
     a_transient_registration?(resource) || a_registration?(resource)
   end
--- a/app/views/shared/registrations/_action_links_panel.html.erb
+++ b/app/views/shared/registrations/_action_links_panel.html.erb
@@ -32,7 +32,7 @@
 
     <% if display_certificate_link_for?(resource) %>
       <li>
-        <%= link_to t(".actions_box.links.certificate"), certificate_link_for(resource) %>
+        <%= link_to t(".actions_box.links.certificate"), registration_certificate_path(resource.reg_identifier) %>
       </li>
     <% end %>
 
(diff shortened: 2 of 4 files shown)


app/helpers/action_links_helper.rb
```diff
@@ -41,6 +41,12 @@ def transfer_link_for(resource)
     new_registration_registration_transfer_path(resource.reg_identifier)
   end
 
+  def certificate_link_for(resource)
+    return "#" unless a_registration?(resource)
+
+    registration_certificate_path(resource.reg_identifier)
```
cintamani: Why creating this method? I think this was useful when the routes where from this system to another, but since this is now inside the same app, can I suggest we get rid of this? We have the `def display_certificate_link_for?(resource)` that will return false when the resource is not correct and hence the link won't appear

irisfaraway: @cintamani I think we have these `x_link_for(resource)` helper methods for all the other links. I like them - I feel like it makes the HTML a lot more readable than having all the jumbled paths, especially as some of them are quite long.

cintamani: No we don't I am quite sure as I got rid of them :D I don't like having an helper method that.... just calls another helper method really! If the path was to be composed I would agree, but here is just unnecessary

irisfaraway: This module has `details_link_for`, `resume_link_for`, `convictions_link_for`, `renew_link_for`, and `transfer_link_for` so I don't think we did get rid of them?

cintamani: And they all needs refactoring ;P to be precise:
- `details_link_for` is still there because we built the show pages as 2 separate routes, we have to refactor it to be under the new `resources` scope, so it will always be one route and we don't need this method any more as there is no condition
- `resume_link_for` is another one I think we should just get rid of, but is arguable that since we have to go through the privacy polici we might want to keep it there for documentation purposes (the method tells me where I will go eventually, even if the path is going to the privacy policy)

irisfaraway: Maybe we just have a difference of opinion on this – I feel like it is a lot easier to read the view when we have the helper methods in place and makes things look cleaner. 🤷‍♀ So I would be leaning towards adding methods for the links that don't already have them.

irisfaraway: Maybe we need @Cruikshanks as a tie-breaker so we can get this story done!

Cruikshanks: pressure

cintamani: 🤣

Cruikshanks: TL;DR I've ended up agreeing with @cintamani

For me, views ideally

- should contain as little code as possible
- should focus on the structure and design of the page and ideally make it clear what I can expect to see when its rendered

On that basis I was originally in agreement with @irisfaraway and could see the benefit of having these helpers to make links clearer. For example `new_registration_registration_transfer_path(resource.reg_identifier)` doesn't jump out to me as well as `transfer_link_for(resource)`!

As I started to document my reasons with examples though, I could see the view already does a good job of making clear what link I'll see rendered. Focusing on `app/views/shared/registrations/_action_links_panel.html.erb` I think there is a pattern being followed that makes it clear to me what is going

```ruby
if display_mawwiage_link_for?(resource)
  linkt_to(t(".actions_box.links.mawwiage"), new_bwings_us_togethew_twue_wove_path(resource.reg_identifier))
end
```

Though the actual Rails generated path helper (obviously my example) isn't very clear, I feel I have everything I need to know that its a link to `mawwiage` I'm going to see rendered. I think it could also be argued its clearer to see in one place that what we render is one thing, but where we actually go to is another.

```ruby
if display_renew_link_for?(resource)
  link_to t(".actions_box.links.renew"), ad_privacy_policy_path(resource.reg_identifier)
end
```

So as long as we are consistent with that pattern, I think we can skip the action link helpers that make the path methods more readable or relevant.

I will now spend my remaining days waiting to hear

> Hello. My name is Iris Faraway. You killed my suggestion. Prepare to die.

irisfaraway: @Cruikshanks 
giphy (1)

Nah, it's cool, I'm happy to go with the majority opinion on this! I'll refactor and then re-request reviews :)

irisfaraway: @cintamani @Cruikshanks Removed the link helper for the certificate here: https://github.com/DEFRA/waste-carriers-back-office/pull/646/commits/0a535035a63a161eec83d83a09f63ebfae4c5311

And I'll look at sorting the others out in a separate PR 👍

PR #653, opened 2020-02-07
Fix finance role permissions

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-884

We need to make a number of fixes to the permissions for our finance roles.

Notably:

- The `finance_user` type exists solely to deal with bank transfer payments
- `finance_super` should be able to do everything a `finance-admin` can do, plus the extra user management bits
- `agency_user`s should not be able to do anything financial

This PR should update all our permissions to match the spreadsheet.

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -47,10 +47,10 @@ class Ability
     permissions_for_agency_user
 
     can :view_revoked_reasons, :all
-    can :refund, :all
     can :cease, WasteCarriersEngine::Registration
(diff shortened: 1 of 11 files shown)


irisfaraway:
...I think the test coverage drop is down to the overall number of lines of code dropping, as we have 100% diff coverage here.

irisfaraway:
This is now ready to review, aside from any charge adjust stuff going on in https://github.com/DEFRA/waste-carriers-back-office/pull/644. Figured we'd deal with that in a separate PR :)

PR #656, opened 2020-02-07
Clean up action link helpers

irisfaraway:
This PR exists to clean up some of the helper methods which we don't think we need any more. This came up during discussion of https://github.com/DEFRA/waste-carriers-back-office/pull/646


irisfaraway:
I've left the following helper methods in:

- `details_link_for` because we haven't refactored to use the new resource URLs yet
- `resume_link_for` and `renew_link_for` because they go through the AD privacy policy and I think it serves a useful documentation purpose

But let me know what you think 👍

PR #664, opened 2020-02-11
Finance - Add ability to revert a payment

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-822
From: https://eaflood.atlassian.net/browse/RUBY-809

This PR contains the code necessary to give a user the ability to revert a payment.
Given the presence of validations, we have created a form for the revert. 
The controller uses our latest standard for dealing with simple journeys in payments, with the exception that we have added an `index` action and route in order to select a payment to be reverted.

### DIFF
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -50,6 +50,12 @@ module ActionLinksHelper
     can?(:refund, resource)
   end
 
+  def display_revert_link?
+    return false unless current_user
+
+    %w[agency_with_refund agency_super finance_admin finance finance_super].include?(current_user.role)
+  end
+
   def display_finance_details_link_for?(resource)
     # TODO: Temporary - for release only. See: https://eaflood.atlassian.net/browse/RUBY-846
     return false
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -51,9 +51,9 @@ module ActionLinksHelper
   end
 
   def display_revert_link?
-    return false unless current_user
+    roles_with_revert_ability = %w[agency_with_refund agency_super finance_admin finance finance_super]
 
-    %w[agency_with_refund agency_super finance_admin finance finance_super].include?(current_user.role)
+    roles_with_revert_ability.include?(current_user.role)
   end
 
   def display_finance_details_link_for?(resource)
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -50,10 +50,10 @@ module ActionLinksHelper
     can?(:refund, resource)
   end
 
-  def display_revert_link?
-    roles_with_revert_ability = %w[agency_with_refund agency_super finance_admin finance finance_super]
+  def display_reverse_link?
+    roles_with_reverse_ability = %w[agency_with_refund agency_super finance_admin finance finance_super]
 
-    roles_with_revert_ability.include?(current_user.role)
+    roles_with_reverse_ability.include?(current_user.role)
   end
 
   def display_finance_details_link_for?(resource)
--- a/app/presenters/payment_presenter.rb
+++ b/app/presenters/payment_presenter.rb
@@ -11,6 +11,10 @@ class PaymentPresenter < WasteCarriersEngine::BasePresenter
     refunded_payment.present?
   end
 
+  def already_reverted?
+    reverted_payment.present?
+  end
+
   def refunded_message
     if worldpay?
       I18n.t(".refunds.refunded_message.card", refund_status: refunded_payment.world_pay_payment_status)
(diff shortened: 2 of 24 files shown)


app/helpers/action_links_helper.rb
```diff
@@ -50,6 +50,12 @@ def display_refund_link_for?(resource)
     can?(:refund, resource)
   end
 
+  def display_revert_link?
+    return false unless current_user
```
irisfaraway: I don't think we need to check for this – if no user is logged in, the user will be redirected anyway.

app/presenters/payment_presenter.rb
```diff
@@ -19,9 +23,28 @@ def refunded_message
     end
   end
 
+  def no_action_message
+    if already_reverted?
+      I18n.t(".reversal_forms.index.already_reversed")
```
irisfaraway:
```suggestion
      I18n.t(".reversal_forms.index.already_reverted")
```

Nitpicking here but it'd be good to standardise on `reverted` or `reversed` 👍 Looks like `reverted` is more common

spec/presenters/payment_presenter_spec.rb
```diff
@@ -52,6 +53,103 @@
...
 
+  describe "#already_reverted?" do
+    before do
+      scope = double(:scope)
+
+      expect(finance_details).to receive(:payments).and_return(scope)
+      expect(scope).to receive(:where).with(order_key: "123_REVERSAL").and_return([reverted_payment])
+    end
+
+    context "if a rerversal payment exist for that order key" do
```
irisfaraway:
```suggestion
    context "if a reversal payment exist for that order key" do
```

spec/presenters/payment_presenter_spec.rb
```diff
@@ -52,6 +53,103 @@
...
+      let(:reverted_payment) { double(:reverted_payment) }
+
+      it "returns true" do
+        expect(subject).to be_already_reverted
+      end
+    end
+
+    context "if a rerversal payment do not exist for that order key" do
```
irisfaraway:
```suggestion
    context "if a reversal payment do not exist for that order key" do
```

spec/presenters/payment_presenter_spec.rb
```diff
@@ -52,6 +53,103 @@
...
+      end
+    end
+
+    context "when the payment was not reverted" do
+      let(:already_reverted) { false }
+
+      it "returns an already reverted message" do
+        result = double(:result)
+
+        expect(I18n).to receive(:t).with(".reversal_forms.index.not_aplicable").and_return(result)
```
irisfaraway:
```suggestion
        expect(I18n).to receive(:t).with(".reversal_forms.index.not_applicable").and_return(result)
```

config/locales/reversal_forms.en.yml
```diff
@@ -0,0 +1,31 @@
...
+      title: "Reverse a payment"
+      heading: "Which payment do you want to reverse?"
+      date: "Date"
+      type: "Type"
+      reference: "Reference"
+      amount: "Amount"
+      action: "Action"
+      reverse_payment: "Reverse payment"
+      reverse_payment_alt: "Reverse payment from %{date} by %{method} for %{amount}"
+      not_aplicable: "n/a"
```
irisfaraway:
```suggestion
      not_applicable: "n/a"
```

app/presenters/payment_presenter.rb
```diff
@@ -19,9 +23,28 @@ def refunded_message
...
+    else
+      I18n.t(".reversal_forms.index.not_aplicable")
```
irisfaraway:
```suggestion
      I18n.t(".reversal_forms.index.not_applicable")
```

app/presenters/payment_presenter.rb
```diff
@@ -19,9 +23,28 @@ def refunded_message
...
+    end
+  end
+
+  def revertible?
```
irisfaraway: Another question about language - we have the `reversible` scope but here we use `revertible`. I assume we are looking for the same conditions, but the different terms could be confusing.

app/helpers/action_links_helper.rb
```diff
@@ -50,6 +50,12 @@ def display_refund_link_for?(resource)
...
+
+    %w[agency_with_refund agency_super finance_admin finance finance_super].include?(current_user.role)
```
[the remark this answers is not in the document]
cintamani: The ability is bounded with the type of payment and we don't have it when we deal with this check. Open to suggestions :)

app/helpers/action_links_helper.rb
cintamani: good point! :D

app/helpers/action_links_helper.rb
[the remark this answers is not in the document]
irisfaraway: I'd be up for making an additional generic `:view_revert_page` ability or something like that. Kind of like how we deal with the list of back office users. We have two types of users who can view that page, but they each have different abilities to actually manage the users.

cintamani:
About `reverse` vs `revert`, I don't think we need to involve content. The wireframe is consistent with the old app and `reverse` is used there, hence I have updated this PR to use `reverse` consistently :) Thanks for the dictionary definitions! This is one of those "English is not my first language" 🙇‍♀

app/helpers/action_links_helper.rb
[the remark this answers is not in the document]
cintamani: I was thinking about that, but then we will have an ability defined for the sake of a view, and it feels a bit wrong to me. Anyway, we have the same problem for payments now (we show the link to anyone as we can't check one ability) so I was thinking let's discuss this case on slack and fix them both in another PR

irisfaraway:
@cintamani I wasn't exactly sure of the difference myself, so no worries at all there :)

app/helpers/action_links_helper.rb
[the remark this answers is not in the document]
irisfaraway: If we have an ability for it then we can also check the authorization on the `index` view, so it would serve multiple uses. Happy to discuss more on Slack though.

PR #669, opened 2020-02-12
Fix view certificate for lower tier registrations

Cruikshanks:

We have recently worked on adding the ability to see the certificate we generate when a registration is complete from the back office. However in QA it was found that an error occurs when you try and use the link for a lower tier registration.

Upon investigation we found this error in the logs

```bash
I, [2020-02-12T14:26:56.768579 #27414]  INFO -- :   Rendered /srv/ruby/waste-carriers-back-office/shared/bundle/ruby/2.4.0/bundler/gems/waste-carriers-engine-e39f56f6bac0/app/views/waste_carriers_engine/pdfs/certificate.html.erb (3.8ms)
I, [2020-02-12T14:26:56.769213 #27414]  INFO -- : Completed 500 Internal Server Error in 11ms
F, [2020-02-12T14:26:56.776346 #27414] FATAL -- :
ActionView::Template::Error (undefined method `in_time_zone' for nil:NilClass):
    125:                 </tr>
    126:                 <tr>
    127:                   <td class="table-column-text"><%=t ".expiry_date" %></td>
    128:                   <td><%= @presenter.expires_on.in_time_zone("London").to_date %></td>
    129:                 </tr>
    130:               </tbody>
    131:             </table>
  app/controllers/certificates_controller.rb:10:in `show'
```

It looks like the current code is always assuming an expiry date will be present which is not the case for lower tier registrations (they never expire). So this change will focus on updating the view not to expect that property.


Cruikshanks:
Don't actually need this as the fix is encapsulated in changes to the engine. See https://github.com/DEFRA/waste-carriers-engine/pull/670

PR #670, opened 2020-02-12
Redirect to correct resource page after renewal is completed

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-895

We have realised that once a renewal have been completed we have the need to redirect the user to the correct resource page, as the transient resource will be deleted at that point.
This contains the refactoring and fix necessary to make that happen.


app/controllers/concerns/can_redirect_form_after_renew.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module CanRedirectFormAfterRenew
+  extend ActiveSupport::Concern
+
+  included do
+    private
+
+    def redirect_after_successful_submit
+      if renew_if_possible
```
Cruikshanks: Just wondering about whether this needs to be in its own concern?

I'm assuming we will only ever use this in conjunction with `CanRenewIfPossible`, and that in most cases where `CanRenewIfPossible` will be used we'll want to redirect afterwards anyway.

So is it worth just adding `redirect_after_successful_submit()` to `CanRenewIfPossible`?

irisfaraway:
I was curious about the coverage drop, so ran the tests locally and I think it's line 28 of the WriteOffFormsController:

```
      if renew_if_possible
        redirect_to resource_finance_details_path(@resource.registration._id)
      else
        redirect_to resource_finance_details_path(@resource._id)
      end
```

We aren't covering a scenario where `renew_if_possible` is false. Is it worth adding one?

spec/requests/worldpay_missed_payment_forms_spec.rb
```diff
@@ -75,8 +75,11 @@
       end
 
       it "redirects to the transient_registration page" do
```
irisfaraway:
```suggestion
      it "redirects to the registration finance page" do
```

spec/requests/cheque_payment_forms_spec.rb
```diff
@@ -75,8 +75,11 @@
       end
 
       it "redirects to the transient_registration page" do
```
irisfaraway:
```suggestion
      it "redirects to the registration finance page" do
```

spec/requests/postal_order_payment_forms_spec.rb
```diff
@@ -75,8 +75,11 @@
       end
 
       it "redirects to the transient_registration page" do
```
irisfaraway:
```suggestion
      it "redirects to the registration finance page" do
```

spec/requests/cash_payment_forms_spec.rb
```diff
@@ -83,7 +83,7 @@
 
         transient_registration.reload
 
-        expect(response).to redirect_to(resource_finance_details_path(transient_registration._id))
+        expect(response).to redirect_to(resource_finance_details_path(transient_registration.registration._id))
```
irisfaraway: Total nitpick, but just wondering if there's a reason why we do this test differently to the other payment ones, where we assign `registration = transient_registration.registration` and then use that in the route?

spec/requests/bank_transfer_payment_forms_spec.rb
```diff
@@ -75,8 +75,11 @@
       end
 
       it "redirects to the transient_registration page" do
```
irisfaraway:
```suggestion
      it "redirects to the registration finance page" do
```

app/controllers/concerns/can_redirect_form_after_renew.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module CanRedirectFormAfterRenew
```
irisfaraway: I'm a little confused about when this is needed – don't we already have this in the ResourceController?

app/controllers/worldpay_missed_payment_forms_controller.rb
```diff
@@ -35,6 +31,16 @@ def authorize_user
     authorize! :record_worldpay_missed_payment, @resource
   end
 
+  def redirect_after_successful_submit
```
Cruikshanks: I wanted to make some comment about how it's a shame we essentially have this code in the concern, so why not use the concern. But then I tried and couldn't make it work well.

So this is more a pointer to me to discuss this bit face-to-face.

cintamani:
I am not happy about this code either 😩Let's get together and discuss it face to face so we might come out with a better solution! Is a tricky one this one!

spec/requests/cash_payment_forms_spec.rb
cintamani: In the other tests, we do that because calling the `let` variable after the object has been deleted (so after the request)  causes the object to be recreated and inconsistent test results. Is a bug we have discussed some months ago. In this case, we are stubbing out the renew job to actually never run, hence we never delete the transient_registration, hence I don't need to ensure that I never call the let variable after the object have been deleted by the request. Is one of those "I have to put a dot at the refactoring somewhere" so this will need more attention later. Changing this test to follow the common pattern means re-writing a lot of it.

cintamani:
I consider Hakiri a false positive, especially because we validate the content of the param before sending it.

spec/requests/cash_payment_forms_spec.rb
cintamani: NM that, after all I have made the refactoring!

app/controllers/concerns/can_renew_if_possible.rb
```diff
@@ -4,18 +4,29 @@ module CanRenewIfPossible
...
 
-      renewal_completion_service = WasteCarriersEngine::RenewalCompletionService.new(@resource)
       renewal_completion_service.complete_renewal
+
+      true
     rescue StandardError => e
       Airbrake.notify(e, reg_identifier: @resource.reg_identifier)
       Rails.logger.error e
+
+      false
```
cintamani: Codeclimate is complaining about this path not being test-covered, but I am not going to add tests here as it would be a lot of code for very little gains.

Cruikshanks: I can live with that. This rescue is part of efforts to understand why things get `STUCK` and so may come out in the future anyway.

cintamani:
They are a constant in my PRs 😂 Sometimes I think I am dyslexic 🤦‍♀I need to create a Grammarly plugin for Sublime!

PR #674, opened 2020-02-13
Fix bug on worlpday missed payment for registrations

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-826

We used to run a little piece of code that would fix the workflow state of a blocked renewal when adding a worldpay missed payment. That code will not work when run on a registration (nor there is the need to run it there). Hence, this PR adds a check that will make sure we only run that code with transient objects which do have a workflow state to work with.

### DIFF
--- a/spec/requests/worldpay_missed_payment_forms_spec.rb
+++ b/spec/requests/worldpay_missed_payment_forms_spec.rb
@@ -72,7 +72,7 @@ RSpec.describe "WorldpayMissedPaymentForms", type: :request do
         sign_in(user)
       end
 
-      it "redirects to the transient_registration page, creates a new payment and assigns the correct updated_by_user to the payment" do
+      it "redirects to the transient_registration's finance details page, creates a new payment and assigns the correct updated_by_user to the payment" do
         old_payments_count = transient_registration.finance_details.payments.count
 
         post "/bo/resources/#{transient_registration._id}/payments/worldpay-missed", worldpay_missed_payment_form: params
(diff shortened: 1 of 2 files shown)


spec/requests/worldpay_missed_payment_forms_spec.rb
```diff
@@ -74,20 +72,26 @@
...
+        expect(response).to redirect_to(resource_finance_details_path(transient_registration._id))
         expect(transient_registration.reload.finance_details.payments.count).to eq(old_payments_count + 1)
+        expect(transient_registration.reload.finance_details.payments.first.updated_by_user).to eq(user.email)
       end
 
-      it "assigns the correct updated_by_user to the payment" do
-        post "/bo/resources/#{transient_registration._id}/payments/worldpay-missed", worldpay_missed_payment_form: params
-        expect(transient_registration.reload.finance_details.payments.first.updated_by_user).to eq(user.email)
+      context "when the resource is a registration" do
+        it "redirects to the transient_registration page, creates a new payment and assigns the correct updated_by_user to the payment" do
```
Cruikshanks: Should that be

```suggestion
        it "redirects to the registration's finance details page, creates a new payment and assigns the correct updated_by_user to the payment" do
```

spec/requests/worldpay_missed_payment_forms_spec.rb
```diff
@@ -74,20 +72,26 @@
         sign_in(user)
       end
 
-      it "redirects to the transient_registration page" do
-        post "/bo/resources/#{transient_registration._id}/payments/worldpay-missed", worldpay_missed_payment_form: params
-        expect(response).to redirect_to(resource_finance_details_path(transient_registration._id))
-      end
-
-      it "creates a new payment" do
+      it "redirects to the transient_registration page, creates a new payment and assigns the correct updated_by_user to the payment" do
```
Cruikshanks: This is more about being consistent with the other change I've suggested (which I think is needed because the current description is incorrect)

```suggestion
      it "redirects to the transient_registration's finance details page, creates a new payment and assigns the correct updated_by_user to the payment" do
```

PR #684, opened 2020-02-16
[SUGGESTION] Simple validation forms suggestion 1

cintamani:
As per conversation with the devs, this is one suggestion of simple forms for validations only

### DIFF
--- a/app/forms/charge_adjust_form.rb
+++ b/app/forms/charge_adjust_form.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class ChargeAdjustForm
+  include ActiveModel::Model
+
+  attr_accessor :charge_type
+
+  validates :charge_type, presence: true
+
+  def submit(params)
+    # Assign the params for validation
+    self.charge_type = params[:charge_type]
+
+    valid?
+  end
+end
--- a/app/controllers/charge_adjust_forms_controller.rb
+++ b/app/controllers/charge_adjust_forms_controller.rb
@@ -5,14 +5,15 @@ class ChargeAdjustFormsController < ApplicationController
 
   prepend_before_action :authenticate_user!
   before_action :authorize_user
+  before_action :fetch_form
 
   def new; end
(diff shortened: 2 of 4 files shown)


app/forms/charge_adjust_form.rb
```diff
@@ -0,0 +1,16 @@
...
+
+  attr_accessor :charge_type
+
+  validates :charge_type, presence: true
+
+  def submit(params)
+    # Assign the params for validation
+    self.charge_type = params[:charge_type]
+
+    valid?
```
cintamani: No need for this?

Cruikshanks:
Having compared this with #685 I've decided this is the solution to opt for.

Solution 3 retains the pattern we have previously implemented but I think its reliance on inheriting from `ResourceFormsController` means it brings in a lot of functionality it doesn't need.

I then did suggest to @cintamani , what if we refactored `ResourceFormsController` to a simpler base that `ResourceFormsController` builds on? She rightly pointed out we already have done that with the concerns. The functionality we need already sits in concerns that `ResourceFormsController ` also pulls in.

I think there could be a case for a `SimpleFormsController` that brought in those shared concerns. It would give a clear name for the type of page a developer is working on and some consistency across them.

But those, like the example this suggestion is built, on are so simple, I think it is more likely to run the risk of premature optimization.

So its a big 👍 and ❤️ for this as our way forward.

irisfaraway:
I'm happy to go with option 1 as well.

I think I agree with what @Cruikshanks said about a `SimpleFormsController` (and a `SimpleForm` as well) - if you open up a file and can immediately tell what type of form or controller it is based on the subclassing, it might make things clearer, particularly as we now have so many forms. But also if it starts to tie us back into knots then it's not worth it.

cintamani:
It might be clearer, but I am unsure about it's maintainability. Let's keep that chat for the next ☕️ and 🍰 meeting :) 

Going to close this PRs now :)

PR #685, opened 2020-02-16
[SUGGESTION] Simple validation forms suggestion 3

cintamani:
As per conversation with the devs, this is one suggestion of simple forms for validations only
In this version, we are adding an `initialize` method which accept a param on the simple form. This gives us the ability to reuse our FormController interfaces.

### DIFF
--- a/app/forms/charge_adjust_form.rb
+++ b/app/forms/charge_adjust_form.rb
@@ -8,7 +8,7 @@ class ChargeAdjustForm
   validates :charge_type, presence: true
 
   # This is here so that we can take advantage of our FormController interfaces
-  def initialize(_resource); end;
+  def initialize(_resource); end
 
   def submit(params)
     # Assign the params for validation
--- a/app/forms/charge_adjust_form.rb
+++ b/app/forms/charge_adjust_form.rb
@@ -7,6 +7,9 @@ class ChargeAdjustForm
 
   validates :charge_type, presence: true
 
+  # This is here so that we can take advantage of our FormController interfaces
+  def initialize(_resource); end;
+
   def submit(params)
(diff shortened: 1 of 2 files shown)


app/forms/charge_adjust_form.rb
```diff
@@ -7,6 +7,9 @@ class ChargeAdjustForm
 
   validates :charge_type, presence: true
 
+  # This is here so that we can take advantage of our FormController interfaces
+  def initialize(_resource); end;
```
cintamani:
```suggestion
  def initialize(_resource); end
```

Cruikshanks:
Having compared this with #684 I have opted for solution 1 over this.

Here is the reason I gave as a comment in #684 

> Having compared this with #685 I've decided this is the solution to opt for.
>
> Solution 3 retains the pattern we have previously implemented but I think its reliance on inheriting from `ResourceFormsController` means it brings in a lot of functionality it doesn't need.
>
> I then did suggest to @cintamani , what if we refactored `ResourceFormsController` to a simpler base that `ResourceFormsController` builds on? She rightly pointed out we already have done that with the concerns. The functionality we need already sits in concerns that `ResourceFormsController ` also pulls in.
>
> I think there could be a case for a `SimpleFormsController` that brought in those shared concerns. It would give a clear name for the type of page a developer is working on and some consistency across them.
>
> But those, like the example this suggestion is built, on are so simple, I think it is more likely to run the risk of premature optimization.
>
> So its a big 👍 and ❤️ for this as our way forward.

cintamani:
Leaving this around for Iris to benefit from this conversation :) will close it when she had an opportunity to take a look

PR #689, opened 2020-02-17
Fix validation on amounts

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-894

This PR fixes some insconsistency on the way we validate our amounts in the payments and charge adjust forms.
This also fixes a bug for which submitting a payment form without an amount results in a 500 error. This bug was dues
to the fact that we were not validating the form data before creating a payment on the object when submitting the form
.
Another issue related to this has to deal with an existing bug on the re-implementation of the Enumerable module from
rails, for which the `sum` method called on a `[nil]` collection returns 0 rather than throwing an error as per ruby's implementation (https://github.com/rails/rails/issues/24796). Given this,
our test suite was missing to report the error because our test always run on objects with no existing payments.

### DIFF
--- a/spec/requests/cheque_payment_forms_spec.rb
+++ b/spec/requests/cheque_payment_forms_spec.rb
@@ -4,8 +4,9 @@ require "rails_helper"
 
 RSpec.describe "ChequePaymentForms", type: :request do
   let(:transient_registration) do
-    create(:renewing_registration, :has_finance_details, :does_not_require_conviction_check)
+    create(:renewing_registration, :does_not_require_conviction_check)
   end
+
   let(:registration) do
     WasteCarriersEngine::Registration.where(reg_identifier: transient_registration.reg_identifier).first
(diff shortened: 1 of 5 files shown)


spec/requests/cheque_payment_forms_spec.rb
```diff
@@ -4,8 +4,9 @@
 
 RSpec.describe "ChequePaymentForms", type: :request do
   let(:transient_registration) do
-    create(:renewing_registration, :has_finance_details, :does_not_require_conviction_check)
```
cintamani: A standard `renewing_registration` object is currently created with existing payments. By using the trait, we were overriding the default behaviour by creating empty finance details. In order to reproduce the issue reported in the ticket, given https://github.com/rails/rails/issues/24796, I have removed the specific call to the trait so that we run our code on a collection that also contains other payments.

PR #693, opened 2020-02-17
Add validation to payments form

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-894

This adds validation to the payment form as per https://github.com/DEFRA/waste-carriers-back-office/pull/684


spec/forms/payment_form_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe PaymentForm do
+  describe "#submit" do
+    context "when params are valid" do
+      let(:params) {{ payment_type: "foo" }}
```
cintamani:
```suggestion
      let(:params) { { payment_type: "foo" } }
```

PR #701, opened 2020-02-18
Fix charge adjustment flash messages

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-870

We spotted during testing that the success message that gets displayed when a charge adjustment is completed is missing the £ symbol and contains a spelling mistake.


cintamani:
Oh! Nice stuff! I had fixed it in another branch too :D

PR #724, opened 2020-02-28
Hide edit functionality behind a feature toggle

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-926

We have implemented feature toggle functionality in the [waste-carriers-engine](https://github.com/DEFRA/waste-carriers-engine/pull/697) which gives us a state we can check in one place, and control through config.

This is to allow us to hide functionality that is not yet ready for release in production, but access it in our non-production environments.

One such feature that is still being developed is the edit registration feature. So this change takes advantage of the new feature toggles, and hides the edit links if the feature is disable.

### DIFF
--- a/config/feature_toggles.yml
+++ b/config/feature_toggles.yml
@@ -1,3 +1,3 @@
 ---
 edit_registration:
-  active: <%= ENV["FEATURE_TOGGLE_EDIT_REGISTRATION"] || "true" %>
+  active: <%= ENV["FEATURE_TOGGLE_EDIT_REGISTRATION"] %>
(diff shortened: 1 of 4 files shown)


Cruikshanks:
As the feature toggle line is only temporary I was going to mark the codeclimate issue as accepted unless anyone has any other thoughts.

config/feature_toggles.yml
```diff
@@ -0,0 +1,3 @@
+---
+edit_registration:
+  active: <%= ENV["FEATURE_TOGGLE_EDIT_REGISTRATION"] || "true" %>
```
irisfaraway: I'd be tempted to set this to FALSE by default as then we don't have to make any env var changes on production.

cintamani: The FeatureToggle class will return false if the env variable is not set, so we just need to remove this `|| true`

PR #764, opened 2020-03-18
Complete a registration if possible at write off

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-954

This fixes a bug for which we fail to complete a registration when a balance is written off from finance users.

### DIFF
--- a/spec/forms/write_off_form_spec.rb
+++ b/spec/forms/write_off_form_spec.rb
@@ -15,16 +15,10 @@ RSpec.describe WriteOffForm do
     context "when the object is valid" do
       let(:valid) { true }
 
-      it "runs the write off service" do
+      it "returns true" do
         comment = double(:comment)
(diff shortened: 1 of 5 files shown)


spec/forms/write_off_form_spec.rb
```diff
@@ -15,24 +15,18 @@
...
         comment = double(:comment)
 
-        expect(ProcessWriteOffService).to receive(:run).with(
-          finance_details: renewing_registration.finance_details,
-          user: user,
-          comment: comment
-        )
-
-        expect(subject.submit({ comment: comment }, user)).to eq(true)
+        expect(subject.submit({ comment: comment })).to eq(true)
```
cintamani:
```suggestion
        expect(subject.submit(comment: comment)).to eq(true)
```

PR #800, opened 2020-04-07
Add api endpoint for acceptance tests

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-977

Add API endpoint to fetch id and token of registrations and renewals. Those will be used by the acceptance tests suite in order to visit pages directly.

### DIFF
--- a/spec/requests/api/registrations_spec.rb
+++ b/spec/requests/api/registrations_spec.rb
@@ -1,0 +1,31 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Assisted Digital Forms", type: :request do
+  let(:registration) { create(:registration) }
+
+  describe "GET /bo/api/registrations/:reg_identifier" do
+    context "when a user is signed in" do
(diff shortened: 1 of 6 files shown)


spec/requests/api/registrations_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Assisted Digital Forms", type: :request do
```
irisfaraway, line 5:
```suggestion
RSpec.describe "Registrations API", type: :request do
```

spec/requests/api/renewals_spec.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Assisted Digital Forms", type: :request do
```
irisfaraway, line 5:
```suggestion
RSpec.describe "Renewals API", type: :request do
```

PR #811, opened 2020-04-16
Add expired registrations in grace window to EPR

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-982 and https://eaflood.atlassian.net/browse/RUBY-983

This changes the scope for fetching registration for the EPR export to include EXPIRED registrations in the grace window.
It also changes the exported data, so that the expire date is inflated with the grace window value for all expired registrations exported.

### DIFF
--- a/app/models/reports/epr_serializer.rb
+++ b/app/models/reports/epr_serializer.rb
@@ -27,7 +27,10 @@ module Reports
     private
 
     def scope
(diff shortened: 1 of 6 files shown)


cintamani:
I really don't know what's SonarCloud problem. The file he reports as missing coverage is actually covered on all red lines. So I am going to ignore it 🤷‍♀️

PR #824, opened 2020-04-22
Add API endpoint to seed data using JSON seeds

cintamani:
This will be used to improve performances of the Acceptance test suite and will make existing seeds file easely reusable.
The main change in seeding is that this API will generate a new registration identifier and return it as a response, so that the test suite can run multiple times without having to wipe the DB data.

Check also: https://github.com/DEFRA/waste-carriers-acceptance-tests/compare/seed-data?expand=1

### DIFF
--- a/app/controllers/api/registrations_controller.rb
+++ b/app/controllers/api/registrations_controller.rb
@@ -2,12 +2,30 @@
 
 module Api
   class RegistrationsController < ApplicationController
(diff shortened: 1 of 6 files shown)


cintamani:
I have no idea what is wrong with SonarCloud, all the lines he claims are not test covered, are actually covered by both unit and request tests

PR #856, opened 2020-05-11
Add cancel functionality for in-progress registrations

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-1017

These implements cancel functionality for in-progress registrations. It has been implemented using the `resources` pattern so that it is easier to extend to in-progress renewals later on.

<details>
<summary>
Confirm cancel
</summary>

Screenshot 2020-05-11 at 15 01 35

</details>

<details>
<summary>
Link to functionality</summary>

Screenshot 2020-05-11 at 14 57 46

</details>

 <details>
<summary>
Confirmation</summary>

Screenshot 2020-05-11 at 16 25 59

</details>

### DIFF
--- a/app/controllers/cancels_controller.rb
+++ b/app/controllers/cancels_controller.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class CancelsController < ApplicationController
+  include CanFetchResource
+  include FinanceDetailsHelper
+
+  prepend_before_action :authenticate_user!
+  before_action :authorize_user
+
+  def new; end
+
+  def create
+    @resource.metaData.cancel
+    @resource.save!
+
+    flash[:success] = I18n.t(
+      "cancels.messages.success",
+      reg_identifier: @resource.reg_identifier
+    )
+
+    redirect_to details_path_for(@resource)
+  end
+
+  private
+
+  def authorize_user
+    authorize! :cancel, WasteCarriersEngine::Registration
+  end
+end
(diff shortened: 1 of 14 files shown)


cintamani:
Ignoring sonar cloud warning on code smell.

app/controllers/cancels_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class CancelsController < ApplicationController
+  include CanFetchResource
+  include FinanceDetailsHelper
```
Cruikshanks, line 5: This isn't an issue with the PR. I'm just taking an opportunity to highlight something _I think_ should be refactored and I'm interested in your thoughts.

We've added [FinanceDetailsHelper](https://github.com/DEFRA/waste-carriers-back-office/blob/master/app/helpers/finance_details_helper.rb) to this controller I'm assuming just to access its `details_path_for()` method.

It stands out to me that we are referring to a finance thing, for something that on the face of it doesn't have anything to do with finance.

So should we find ourselves with 5 minutes spare 😁 , I think we should be refactoring `details_path_for()` to a more appropriately named Concern. What do you think?

cintamani: I think we should keep it as a helper because the function is used mostly in views than in controllers. What I suggest is to move the function to a ResourceHelper class at this point. WDYT?

Cruikshanks: If it's used in views more than controllers that makes sense. But don't worry about updating this PR. It should be done in a separate one.

PR #860, opened 2020-05-15
Update templates of first and second renewal email reminders

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-1019

This update the template of the first and second email reminders to match the magic link template. In the new template, we don't have differences for those email templates. Little refactoring has been done there, as it is quite possible that the email templates will differ again in the near future.

Screenshot 2020-05-15 at 14 31 41


irisfaraway:
@cintamani Should this be behind a feature flag? I thought there might be a possibility of us shipping reminders before the magic link renewal was ready.

cintamani:
@irisfaraway this is already behind a feature flag: https://github.com/DEFRA/waste-carriers-back-office/blob/master/lib/tasks/email.rake#L13

irisfaraway:
@cintamani Sending reminders at all is behind a feature flag, yeah. But I meant a flag for sending reminders vs sending out the magic link. Since it looks like this overrides the previous version of the reminder email entirely.

irisfaraway:
Ignore me, I missed a conversation at some point and this is not an issue!

PR #878, opened 2020-05-26
Add view details page for new registrations

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1054

We want NCCC staff to be able to access details of a new in-progress registration, as they can with renewals.

This PR sets up a version of the 'view details' page for in-progress registrations. This page should be linked to in the relevant search results.

We also don't want to include any finance info on this page.

### DIFF
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -20,10 +20,6 @@ module ActionLinksHelper
     ad_privacy_policy_path(reg_identifier: resource.reg_identifier)
   end
 
-  def display_details_link_for?(resource)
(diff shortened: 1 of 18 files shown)


app/helpers/action_links_helper.rb
```diff
@@ -18,7 +21,7 @@ def renew_link_for(resource)
   end
 
   def display_details_link_for?(resource)
-    a_transient_registration?(resource) || a_registration?(resource)
```
cintamani, line 21: we should probably get rid of this method. Whatever the search result returns, will have a detail link now :)

irisfaraway: Agreed. It's a goner!

PR #886, opened 2020-05-28
Add rake task to remove old transient_registrations

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-10

We've agreed that we shouldn't store transient_registrations for more than 30 days. So we need a way of clearing out the old ones.

This rake task will find and remove transient_registrations that are older than 30 days. It should also remove all related objects (eg transient_addresses).


irisfaraway:
I don't think we usually test the rake tasks so that's why code coverage is dropping.

PR #891, opened 2020-05-29
Remove older transient_registrations with no created_at

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-10

When I first wrote this service in https://github.com/DEFRA/waste-carriers-back-office/pull/886 I was planning to deal with the older transient_registrations that don't have a `created_at` value set by just having a one-off script to clean them out.

However, on further thinking, phasing them out will likely be a more gradual process. So it makes more sense to build it into this scheduled task instead.

The added code will find any transient_registrations which don't have a `created_at` and are not submitted renewals, and then delete them based on their `last_modified` date.

### DIFF
--- a/app/services/transient_registration_cleanup_service.rb
+++ b/app/services/transient_registration_cleanup_service.rb
@@ -3,6 +3,7 @@
 class TransientRegistrationCleanupService < ::WasteCarriersEngine::BaseService
   def run
     transient_registrations_to_remove.destroy_all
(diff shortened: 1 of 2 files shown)


irisfaraway:
@cintamani I don't think we can just run it once, which is why I've put it in here as well. At first I was thinking we could just run it when we release it, but actually there will be some which are still in progress so we'd need to run it repeatedly to phase them out. But I think eventually we can remove this code, just not at first.

PR #893, opened 2020-06-02
Don't display resume link for new regs unless users have permission

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1055

Users who can't create new registrations (ie finance users) should not see the 'resume' link for in-progress new registrations.


irisfaraway:
Also... the ActionLinksHelper is starting to get kind of icky. I'd like to do some more refactoring on it but thought I'd tackle that as part of a separate PR.

PR #915, opened 2020-06-18
Upgrade ruby and rails versions

cintamani:
From: https://eaflood.atlassian.net/browse/RUBY-1061

This PR contails all the changes necessary to upgrade the Ruby version from 2.4 to 2.7.1 and Rails version from 4.2 to 6.0.3.1

Some notes about the updates:
- Request tests syntax for params and headers to send along with a `get`, `put` or `post` request have changed. Now the params are into a key-value format with the key specifying what are `params` and what are `headers`
- With Rails 6, the `rake` command is not invoked directly any more. Instead, we now use `bin/rails`.
- Since Rails 6 was built with the. intent to leave the assets work to. NodeJs, we had to require the `sassc-rails` gem on top of the rest of the gems in order for the old way of dealing with assets still works.
- `POST` request that do not specify a render will redirected to a `GET` automatically, hence we have to explicitly raise errors now if we want them not to be reachable
- Inherithance and scoping are something we must make more attention too. We used to define stuff like:
```
module WasteCarriersEngine
  class ErrorsController < ApplicationController
```
and trust that ruby would resolve the `ApplicationController` to be the module-scoped definition of it. We can't trust that any more, hence we have to be specific and do stuff like:
```
module WasteCarriersEngine
  class ErrorsController < ::WasteCarriersEngine::ApplicationController
```
- Parameters on controller coming from a request are not any more instances of the Hash class, but instances of the `ActionController::Parameters` class
- The test suite will run in `:random` mode by default
- Many of the spec matches we were used to in request tests have been externalised to a gem `rails-controller-testing`


config/environments/test.rb
```diff
@@ -15,8 +15,8 @@
   config.eager_load = false
 
   # Configure static file server for tests with Cache-Control for performance.
-  config.serve_static_files   = true
-  config.static_cache_control = "public, max-age=3600"
+  config.public_file_server.enabled = true
+  config.public_file_server.headers = { 'Cache-Control' => 'public, max-age=3600' }
```
cintamani, line 19:
```suggestion
  config.public_file_server.headers = { "Cache-Control" => "public, max-age=3600" }
```

config/environments/development.rb
```diff
@@ -11,9 +12,23 @@
...
+    config.cache_store = :memory_store
+    config.public_file_server.headers = {
+      "Cache-Control" => "public, max-age=172800"
+    }
+  else
+    config.action_controller.perform_caching = false
+
+    config.cache_store = :null_store
+  end
+
```
cintamani, line 31:
```suggestion
```

cintamani:
Hakiri is to be marked as false positive

PR #921, opened 2020-06-24
Amend acceptance test seed API for expires_on

Cruikshanks:
https://github.com/DEFRA/waste-carriers-acceptance-tests/pull/330

Whilst working on adding the first scenarios to the [acceptance-tests](https://github.com/DEFRA/waste-carriers-acceptance-tests) to make use of setting a custom `expires_on` date we encountered an issue.

Prior to this change the API would set the `expires_on` if the seed data it received _didn't_ include the property `"custom_expire"`. But we found in the acceptance tests there was an issue in the logic which meant this was never getting applied.

So we then had to add additional logic to ensure not only was that field added to the seed data, but that when it was we also set the `expires_on` field. Having made this change, we then immediately concluded we were only setting the `expires_on` when we wanted a custom expires date. So why not simplify things both in the acceptance tests and the API and just look for the presence of `expires_on` in the seed data.

If it's there do nothing. Assume whoever wrote the test knows what date they wanted set. If its not there, stick with the current logic of automatically applying one.


Cruikshanks:
Running tests during the 'witching hour' so stuff date related is failing. Will poke again in the morning 😁

PR #948, opened 2020-07-16
Add payment confirmation email to view details

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1159

Now we are collecting a payment confirmation email during the registration and renewal journey we should really be showing it in the view details page. Else there will be no means to see this in the system.

<details><summary>Example</summary>

Screenshot 2020-07-16 at 22 32 42

</details>

### DIFF
--- a/app/views/registrations/show.html.erb
+++ b/app/views/registrations/show.html.erb
@@ -50,6 +50,10 @@
         <% if @registration.main_people.any? %>
           <%= render "shared/registrations/main_people_details", resource: @registration %>
         <% end %>
+
+        <% if @registration.receipt_email.present? && !@registration.receipt_email.empty? %>
+          <%= render "shared/registrations/receipt_email_panel", resource: @registration %>
+        <% end %>
       </div>
 
       <div class="column-one-third">
--- a/app/views/registrations/show.html.erb
+++ b/app/views/registrations/show.html.erb
@@ -51,7 +51,7 @@
           <%= render "shared/registrations/main_people_details", resource: @registration %>
         <% end %>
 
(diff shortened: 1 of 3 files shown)


app/views/registrations/show.html.erb
```diff
@@ -50,6 +50,10 @@
         <% if @registration.main_people.any? %>
           <%= render "shared/registrations/main_people_details", resource: @registration %>
         <% end %>
+
+        <% if @registration.receipt_email.present? && !@registration.receipt_email.empty? %>
```
irisfaraway, line 54:
```suggestion
        <% if @registration.receipt_email.present? %>
```

`.present?` is more thorough so should cover us!

Cruikshanks:
Taking it back to draft. Need to make a quick tweak.

irisfaraway:
Oops, sorry, just saw you returned it to draft!

app/views/registrations/show.html.erb
Cruikshanks: Awesome. Had it in my head it only covered the field being there 🤦

Cruikshanks:
SonarCloud is complaining about the accessibility elements of the table I've now added. It is right, but then we are using the table for layout and not for displaying data.

WCAG advises you don't do this and use CSS instead. However, if you do have to do it they say not to include `caption` and `scope` elements.

[F46: Failure of Success Criterion 1.3.1 due to using th elements, caption elements, or non-empty summary attributes in layout tables](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/F46)

So I think we're just going to have to ignore SonarCloud on this one and ideally look to refactor to use CSS in the future (when we get those magical 5 minutes!)

PR #949, opened 2020-07-16
Add payment confirmation email to Edit

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1160

Now we have the new `receipt_email` field which we are capturing during new registrations and renewals we need to allow users to edit the value. Else once set (or not) it can never be changed which will cause issues for users who need to change it for subsequent Worldpay payments, for example, copy card orders.


Cruikshanks:
Edit functionality is actually in the engine. So hold this through whilst I go work on <https://github.com/DEFRA/waste-carriers-engine/pull/891> 😁

Cruikshanks:
Closing as there is a series of changes, including one related to this I need to bring in.

PR #953, opened 2020-07-20
Extend back office renewal grace period

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1153

Currently, if a user expires and is past the grace window, they have to make a new registration. This means they cannot keep their old reg number and have to pay an additional cost for a new reg (as renewals are cheaper).

NCCC will often choose to refund these users for that extra fee when they are contacted (although they are not obligated to do this). This is an admin-heavy process.

To make this easier for them, we’ve decided to allow NCCC users to always be able to renew an expired reg regardless of the grace window. This will mean they don’t need to deal with as much admin, and the user gets to keep their number.

This should not affect the front office.

---

Depends on https://github.com/DEFRA/waste-carriers-engine/pull/893

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,7 +51,7 @@ gem "defra_ruby_aws", "~> 0.3.0"
 # Use the waste carriers engine for the user journey
 gem "waste_carriers_engine",
     git: "https://github.com/DEFRA/waste-carriers-engine",
-    branch: "main"
+    branch: "expiry-refactor"
 
 # Use the Defra Ruby Features gem to allow users with the correct permissions to
 # manage feature toggle (create / update / delete) from the back-office.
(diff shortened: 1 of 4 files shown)


irisfaraway:
The reason this has "3" in the branch name is it's the third implementation attempt.

Attempt 1 was just fiddling with the application config to read different env vars as in the old PR.

Attempt 2 set up the feature flag in the back office's engine configuration, and then read that config in the engine's ExpiryCheckService. This looked pretty nice but because the config was only loaded when the app started, toggling the flag didn't take effect until the app itself was restarted. So that was out.

So here's Attempt 3 – less pleasing, but we've done it before in WEX and it works!

PR #956, opened 2020-07-21
Add template for non-AD renewal letters

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-255

This PR sets up the template for the non-AD renewal letters, as well as sorting any pending tests.


irisfaraway:
All of these Sonarcloud issues are about the CSS. Honestly I agree that it is a mess but it's what we have in WEX and it works. So I'd rather just leave it as-is rather than start poking it.

PR #962, opened 2020-07-23
Generate renewal token when one is requested

Cruikshanks:

If a registration is eligible for renewal NCCC would like us to display the same renewal link that gets included in the renewal reminder emails and letters in registration details.

Currently though we only generate the token used in the link when first generating the reminder email. A registration is eligible for renewal before this is done. And when we first release there will be hundreds that are eligible which will never go through the reminder process because it is too late.

The simplest solution we have for this is to generate it if required. We made that change in the engine in [PR #898](https://github.com/DEFRA/waste-carriers-engine/pull/898). This change updates the back-office to use the new functionality.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,7 +51,7 @@ gem "defra_ruby_aws", "~> 0.3.0"
 # Use the waste carriers engine for the user journey
 gem "waste_carriers_engine",
     git: "https://github.com/DEFRA/waste-carriers-engine",
-    branch: "generate-renewal-token-on-request"
+    branch: "main"
 
 # Use the Defra Ruby Features gem to allow users with the correct permissions to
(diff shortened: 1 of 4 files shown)


Cruikshanks:
There is an argument to be made following this change that we could refactor the renewal reminder mailer so that there is only the one `reminder_email()` method.

I've left it for now because having the distinction between the 2 does mean there is some flexibility should we need it to have different content in each email. However, wouldn't take much to push me to refactor now and worry about flexibility when we need it.

Cruikshanks:
It was highlighted that this change drops the call to `WasteCarriersEngine::FeatureToggle.active?(:renew_via_magic_link)`. However, PR #964 will be doing away with it anyway so I no longer see it as an issue.

PR #963, opened 2020-07-23
Display renewal reminder magic link in reg. detail

Cruikshanks:

If a registration is eligible for renewal NCCC would like us to display the same renewal link that gets included in the renewal reminder emails and letters in registration details.

However, the link should not be clickable but shown as plain text. This is to avoid a user accidentally clicking on it and starting the renewal.

The reasoning behind it is so NCCC users can manually send the link to someone, for example, a 3rd party agent, who has a different email from the contact email stored in the registration.

<details><summary>Example</summary>

Screenshot 2020-07-24 at 00 35 48

</details>

### DIFF
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -23,8 +23,7 @@ module ActionLinksHelper
   end
 
   def display_renewal_link_for?(resource)
-    return false unless WasteCarriersEngine::FeatureToggle.active?(:renew_via_magic_link)
-    return false unless WasteCarriersEngine::FeatureToggle.active?(:display_renewal_link)
+    return false unless WasteCarriersEngine::FeatureToggle.active?(:renewal_reminders)
 
     display_renew_link_for?(resource)
   end
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -22,6 +22,16 @@ module ActionLinksHelper
     ad_privacy_policy_path(reg_identifier: resource.reg_identifier)
   end
 
+  def renewal_link_for(resource)
+    return "" unless a_registration?(resource)
+
+    [
+      Rails.configuration.wcrs_renewals_url,
+      "/fo/renew/",
+      resource.renew_token
+    ].join
+  end
+
   def display_write_off_small_link_for?(resource)
     can?(:write_off_small, resource) && resource.balance != 0
   end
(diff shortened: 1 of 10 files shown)


Cruikshanks:
Sonarcloud issues are the usual round tables not having accessibility elements. However, as we are using them for layout W3C recommends not including those elements.

Cruikshanks:
Because this was branched off another branch and has then been rebased, there are more commits listed than actually matter making changes to files not listed in **Files changed** ?!

So if working through the commits ignore anything before `Display renewal reminder magic link in reg. detail` 😁

app/helpers/action_links_helper.rb
```diff
@@ -22,6 +22,23 @@ def renew_link_for(resource)
     ad_privacy_policy_path(reg_identifier: resource.reg_identifier)
   end
 
+  def display_renewal_link_for?(resource)
```
irisfaraway, line 25: Can we give this method a more specific name? I feel like `display_renewal_link_for` and `display_renew_link_for` are too similar! Maybe specify this is the magic link?

app/helpers/action_links_helper.rb
```diff
@@ -22,6 +22,23 @@ def renew_link_for(resource)
...
+    return false unless WasteCarriersEngine::FeatureToggle.active?(:renewal_reminders)
+
+    display_renew_link_for?(resource)
+  end
+
+  def renewal_link_for(resource)
```
irisfaraway, line 31: Might be something for a different PR, but as this URL is now being used here and in the letters and emails, would be great if we were just generating it in a single place instead :D

Cruikshanks: Wholeheartedly agree! But I wasn't sure how to go about it. I rely on stuff in the action links helper here which isn't needed in the renewal reminder mailer.

So I couldn't see how I could split it out into something shared, without a wholesale refactoring. That felt too risky at this time. But I would _love_ it if there was some solution I've overlooked or I'm just overthinking what is needed 🤷‍♂️ 😁

app/helpers/action_links_helper.rb
Cruikshanks: Perfectly fair point. I was taking the name from the title of the new section to try and make that link obvious. Bloomin naming things is 💩

app/helpers/action_links_helper.rb
Cruikshanks: Ignore me, being dumb. Got a simple solution for this 🤦

PR #964, opened 2020-07-24
Merge email_reminders and renew_via_magic_link

Cruikshanks:
https://eaflood.atlassian.net/browse/RUBY-1170

Our intention was to provide automated renewal reminders and build on it with adding magic links and being able to renew through them.

So we added added 2 separate feature toggles to control what was enabled.

In hindsight, the app hasn't actually been built to operate for each of the combinations those 2 feature toggles could be in. Essentially now we consider renewals will only ever be shipped with magic link functionality.

So this change simplifies the logic around renewals by merging these toggles into a new `renewal_reminders` feature toggle.

It also encompass changes to ensure renewal reminder content is correctly enabled and disabled depending on the state of the toggle.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -31,7 +31,7 @@
                               main_app.users_path %>
                 </li>
               <% end %>
-              <% if can?(:manage, FinalReminderLettersExport) %>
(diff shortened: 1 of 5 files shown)


Cruikshanks:
Thing to note: I purposefully haven't hidden the actual letter generation behind the `renewal_reminders` feature flag. It does not hurt to still have them generating even if renewal reminders are not enabled.

But having them generating means if and when the feature is enabled, the letters will be there ready and waiting!

PR #971, opened 2020-07-27
Fix broken renewal link panel

Cruikshanks:

In [PR #963](https://github.com/DEFRA/waste-carriers-back-office/pull/963) we added a panel to the registration details to display the renewal magic link. This will allow back-office users to easily copy and paste the link somewhere else, for example if a registration is actually managd by a 3rd party.

As part of that PR some feedback suggested renaming some of the helpers to make things clearer. That change was made but one reference got missed.

This change fixes that, and renames the panel to make the naming consistent.


Cruikshanks:
Sonarcloud issues are the usual round tables not having accessibility elements. However, as we are using them for layout W3C recommends not including those elements.

PR #974, opened 2020-07-27
Fix intermittent test failures for letter exports

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-255
https://eaflood.atlassian.net/browse/RUBY-251

We've spotted that the specs for the letter exports, specifically the ReminderLettersExportService specs, will occasionally fail in the Travis builds.

This PR is for debugging and fixing this issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,5 +58,5 @@ before_script:
 
 script:
   - bundle exec rubocop --format progress --format json --out rubocop-result.json
-  - bundle exec rspec
+  - bundle exec rspec --seed 13731
   - sonar-scanner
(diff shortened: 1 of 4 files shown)


Cruikshanks:
On the basis you have set a seed I assume you can always get it to fail now. So perhaps the order of the tests might be a cause. I know Cinti made changes/felt the database wasn't always cleaning properly between tests. Would forcing a database clean before starting these particular tests help in any way? 🤷‍♂️

irisfaraway:
Unclear what is causing this and we have to move on. So logging an issue and closing this for now https://github.com/DEFRA/ruby-services-team/issues/83

PR #1073, opened 2021-04-06
Remove code to migrate old backend users to back office

irisfaraway:
https://eaflood.atlassian.net/browse/RUBY-1367

This used to be required when we needed to keep the backend and back office in sync. However the backend is now long gone so this code is no longer necessary.


irisfaraway:
Hakiri webhook failed again, so ran the scan manually and it's fine.

PR #1082, opened 2021-04-26
Bump cancancan

tobyprivett:
This a bump from v1 to v3 and would probably benefit from smoke testing.


Gemfile
```diff
@@ -22,7 +22,7 @@ gem "sdoc", "~> 1.1.0", group: :doc
 
 # Use CanCanCan for user roles and permissions
 # Version 2.0 doesn't support Mongoid, so we're locked to an earlier one
```
irisfaraway, line 24:
```suggestion
```

Think we can drop this line now 🎉

tobyprivett: yes! done

irisfaraway (approved):

PR #1083, opened 2021-04-26
Send first reminder emails from Notify

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1281


irisfaraway (commented):
This is looking good! Just some questions about the BaseEmailSendService(s).

app/services/notify/base_send_email_service.rb
```diff
@@ -0,0 +1,28 @@
...
+      @registration = registration
+
+      client = Notifications::Client.new(WasteCarriersEngine.configuration.notify_api_key)
+
+      client.send_email(notify_options)
+    end
+
+    private
+
+    def validate_contact_email(registration)
```
tobyprivett, line 18: As this method is validating AD, it could be best kept within the back-office. If the behaviour is needed by other classes, we could extract to a mix in or service object. For now, though, am thinking it could stay here. Happy to move it elsewhere, if that would be best.

irisfaraway: Back office sounds fine to me 👍

tobyprivett: Thanks!

irisfaraway (approved):

PR #1088, opened 2021-04-28
Re-send confirmation emails from Notify on demand

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1284


irisfaraway (commented):
Love all the refactoring here 😍

Only suggestion (and it is a small one!) is that we have some specific tests for the abilities that could do with having the `can :resend_confirmation_email` added to them. It would go in spec/support/shared_examples/abilities/agency_examples.rb

irisfaraway (approved):
:shipit:

PR #1101, opened 2021-05-07
Send Upper tier AD confirmation letter automatically through Notify

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1286


tobyprivett:
@irisfaraway The first few commits are mostly refactoring. Happy to take a different approach, if that's preferred.

irisfaraway (changes requested):
This is looking good! I appreciate all the separate commits – made it very easy to step through and understand the changes.

However, I think this implementation will only send when the registration is activated later after submission, rather than if the initial submission makes the registration activate immediately. So the letter would get sent once a pending payment or conviction check was resolved, as that's where the back office controllers are completing the process – however if the completion happens in the engine when the forms are submitted, there's nothing there to trigger the letter being sent.

Having the letter code in the engine and triggering it from the RegistrationActivationService or RenewalCompletionService (like the confirmation emails) might get around this issue.

Hope that makes sense – happy to discuss!

tobyprivett:
Understood. In the engine, the RenewalCompletionService wraps the RegistrationActivationService, so letters could be delivered by the RegistrationActivationService. Does that sound like the best approach?

The namespace refactor might be useful down the line, but we can always retrieve the commits later.

I will close this and implement in the engine

PR #1113, opened 2021-05-17
Integrate account transfer email in Notify

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1426

When a registration is transferred, one of two emails is sent:
- For a new user
- For an existing user

Here we replace ActionMailer with Notify to deliver the emails.


tobyprivett:
Closing this as accidentally named the branch `chore/`

PR #1117, opened 2021-05-19
Handle no account_email on registration transfer page

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1429

- An edge case may arise whereby the account_email is blank,
so we need to display alternate text on the registration
transfer page


tobyprivett:
@irisfaraway We seem to be presenter-driven in this app, and I wanted to avoid un-testable switches in the view templates. Happy to approach this in a different way, though.

irisfaraway (approved):

PR #1179, opened 2021-11-11
Delete expired, ceased, revoked registrations after 7 years

tobyprivett:
https://eaflood.atlassian.net/browse/RUBY-1683

We have a requirement to delete old registrations that been ceased, revoked or expired more than seven years ago. 

There are several components to this task:
- [x] An overnight job is scheduled to run this task
- [x] The task removes all registrations from the database which expired more than 7 years ago
- [x] Removes all registration from the database which ceased or revoked more than 7 years ago
- [x]  Support testing with some [registrations that can be deleted ](https://github.com/DEFRA/waste-carriers-front-office/pull/745) to support testing
- [x] This [job can be run through Jenkins](https://wcr-jenkins.aws-int.defra.cloud/job/DEV_77_RUN_REMOVE_DELETABLE_REGISTRATIONS_JOB/) on non-prod environments for test purposes
- [x] Update the relevant documentation: https://github.com/DEFRA/ruby-services-team/blob/main/services/wcr/background_jobs.md

### DIFF
--- a/app/presenters/base_registration_presenter.rb
+++ b/app/presenters/base_registration_presenter.rb
@@ -95,6 +95,10 @@ class BaseRegistrationPresenter < WasteCarriersEngine::BasePresenter
            reg_identifier: reg_identifier)
   end
 
+  def display_last_modifed
+    metaData.last_modified.to_datetime
+  end
+
   private
 
   def displayable_tier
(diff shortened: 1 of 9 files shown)


timstone (approved):
I'm uncertain if old lower tier registrations were classes as inactive until the user clicked on a link in an email to active them, I'm only going to be able to test that once we use prod data on preprod.

PR #1217, opened 2022-01-26
Display a registration's metaData

tobyprivett:
An adhoc requirement that enables NCCC users to view the metaData for a registration.

https://eaflood.atlassian.net/browse/RUBY-1766


tobyprivett:
It feels *wrong* not to accompany this PR with a test, but I can't see a sensible way to add one.

PaulDoyle-DEFRA:
> It feels _wrong_ not to accompany this PR with a test, but I can't see a sensible way to add one.

Possibly add a simple feature spec to confirm that the metadata section is present and contains at least one attribute value?

PaulDoyle-DEFRA (approved):

PR #1230, opened 2022-02-09
Update AD privacy notice page

tobyprivett:
- Here we link to the front office privacy policy page,
rather than rendering inline
- This is to avoid duplicating the text

https://eaflood.atlassian.net/browse/RUBY-1718


config/locales/ad_privacy_policy.en.yml
```diff
@@ -5,26 +5,11 @@ en:
       heading: "Privacy policy for Waste Carriers, Brokers and Dealers assisted digital"
       sub_heading: "Tell the customer:"
       privacy_policy: "privacy policy"
-      first_paragraph: "The %{link_to_privacy_policy} for the Waste Carriers, Brokers and Dealers Service has been updated, and can be viewed online at: wastecarriersregistration.service.gov.uk/fo/pages/privacy"
+      first_paragraph: "The privacy policy for the Waste Carriers, Brokers and Dealers Service has been updated, and can be viewed online at: wastecarriersregistration.service.gov.uk/fo/pages/privacy"
```
PaulDoyle-DEFRA, line 8: This is possibly outside the scope of the change, but shouldn't the piece after "online at:" be an actual URL?

PR #1232, opened 2022-02-10
Feature/card order export registration number

PaulDoyle-DEFRA:
This change resolves three issues with the content of the cards export CSV file:
1. If address line 1 in the database matches the business name, that address line should be skipped to avoid duplication of the business name on the card.
2. The value in the Registration Number column should be the registration identifier, not the registration (database) ID.
3. Dates should be presented in dd/mm/yy format, without the time element. (Note that the date format is to be confirmed with the stakeholders and may need to be updated).

### DIFF
--- a/app/presenters/card_orders_export_presenter.rb
+++ b/app/presenters/card_orders_export_presenter.rb
@@ -23,7 +23,7 @@ class CardOrdersExportPresenter < WasteCarriersEngine::BasePresenter
   end
 
   def reg_identifier
-    @order_item_log.registration_id
+    @registration.reg_identifier
   end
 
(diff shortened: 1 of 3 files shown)


PaulDoyle-DEFRA:
> For the sonar complaints, we could use one of these date/time formats (or add another): https://github.com/DEFRA/waste-carriers-back-office/blob/main/config/initializers/time_formats.rb

The current report format is US-style m/d/yy, which is understandably not included in that set of formats. Awaiting clarification from the stakeholder on the preferred date format and will revisit this when clarified.

tobyprivett (approved):

PR #1236, opened 2022-02-11
Feature/card export orders page

PaulDoyle-DEFRA:
This change adds a back-office page to view copy card order exports. It also tracks and displays the user and time of the first click on the download link for each export.
Ref. https://eaflood.atlassian.net/browse/RUBY-110


lib/tasks/export_copy_card_orders.rake
```diff
@@ -7,7 +7,7 @@ namespace :reports do
 
       # Run the report for one week up to end of day on the most recent Sunday
```
irisfaraway, line 8: Does this comment need updating? Or was it just wrong before?

app/views/card_order_exports/index.html.erb
```diff
@@ -0,0 +1,49 @@
...
+    <% if @exports.last.present? %>
+      <h2 class="govuk-heading-m">The latest file was created at <%= @exports.last.exported_at %></h2>
+    <% end %>
+  </div>
+</div>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-three-quarters">
+
+    <table class="govuk-table" aria-label="<%= t(".heading") %> >
```
irisfaraway, line 17: Looks like Sonarcloud would like a `caption` or `aria-describedby` for this. Can we check with Will for text?

irisfaraway (commented):
This is looking good, nice and clear! Hopefully the CI issues are fairly quick to squash.

lib/tasks/export_copy_card_orders.rake
PaulDoyle-DEFRA: Yes, it needs to be updated.

app/views/card_order_exports/index.html.erb
PaulDoyle-DEFRA: I added a caption, will check with Will on the wording.

irisfaraway (approved):

PR #1277, opened 2022-03-16
Display `entity_display_name` in search results

tobyprivett:
Here we introduce a `SearchResultPresenter` that mixes in `CanPresentEntityDisplayName` so that the `entity_display_name` can be displayed in the search results.

Unfortunately, the ActionLinksHelper uses `is_a?(model-klass)` to
make page display decisions. To support this behaviouer, we enable that method to be called 
on the model that was passed to this class.

This is an unfortunate clash of design patterns - if there is a better solution, please let me know!

https://eaflood.atlassian.net/browse/RUBY-1794

### DIFF
--- a/spec/presenters/search_result_presenter_spec.rb
+++ b/spec/presenters/search_result_presenter_spec.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe SearchResultPresenter do
+  let(:presenter) { described_class.new(model) }
+
+  describe "#entity_display_name" do
+    let(:model) { Object.new }
+
+    it { expect(presenter).to respond_to(:entity_display_name) }
+  end
+
+  describe "#is_a? WasteCarriersEngine::NewRegistration" do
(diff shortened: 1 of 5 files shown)


spec/presenters/search_result_presenter_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+RSpec.describe SearchResultPresenter do
+  let(:presenter) { described_class.new(model) }
+
+  describe "#entity_display_name" do
+    let(:model) { Object.new }
+
+    it { expect(presenter).to respond_to(:entity_display_name) }
+  end
+
+  describe "#is_a? WasteCarriersEngine::NewRegistration" do
```
Beckyrose200, line 14: What does the # at the beginning mean? Is it to do with testing a method? :)

tobyprivett: It's an rspec convention. For an instance method we could write `describe "#instance_method_name"` and for a class method `describe ".class_method_name"`

PR #1307, opened 2022-04-07
Refactor worldpay card payments

PaulDoyle-DEFRA:
These changes are a preparatory step for migration from Worldpay to GovPay. The aim is to separate general online-payments logic from Worldpay-specific logic.
Changes include:
- Rename classes referencing “worldpay” which are about online payments in general and not worldpay specifically.
- Rename keys (not values) which reference worldpay in locale files.
- Changes to the following are specifically avoided:
- Names of worldpay-specific classes.
- Database schema (implied) or contents - i.e. the database still has contents such as payment_status: "WORLDPAY_MISSED".
- Environment settings
- Locale file values (as opposed to keys)
- Anticipatory refactoring based on assumptions about how GovPay integration will be implemented.

https://eaflood.atlassian.net/browse/RUBY-1848

### DIFF
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -88,7 +88,7 @@ class Ability
     can :view_payments, :all
 
     can :reverse, WasteCarriersEngine::Payment do |payment|
-      payment.worldpay? || payment.worldpay_missed?
+      payment.online? || payment.online_missed?
     end
   end
 
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -33,7 +33,7 @@ class Ability
   # Permissions for specific roles
 
   def permissions_for_agency_user
-    # This covers everything mounted in the engine and used for the assisted digital journey, including WorldPay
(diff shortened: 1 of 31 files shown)


app/models/ability.rb
```diff
@@ -84,11 +84,11 @@ def permissions_for_finance_admin_user
     can :charge_adjust, :all
     can :write_off_large, WasteCarriersEngine::FinanceDetails
     can :view_certificate, WasteCarriersEngine::Registration
-    can :record_worldpay_missed_payment, :all
+    can :record_online_missed_payment, :all
```
tobyprivett, line 87: Previously `online_payment` has been directly substituted for `worldpay`. I agree that `record_online_payment_missed_payment` might be a bit clumsy - but is it clearer?

PaulDoyle-DEFRA: Fair point, for consistency this should be aligned with the use of `online_payment` elsewhere. Although `payment` is already part of the name. Perhaps `record_online_payment_missed` would be both more consistent and less clumsy?

PR #1314, opened 2022-04-12
Enable mock mode for Rails.config.companies_house_host

tobyprivett:
- Rails.config.companies_house_host is used in the DefraRubyCompaniesHouse class

https://eaflood.atlassian.net/browse/RUBY-1855

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -71,8 +71,14 @@ module WasteCarriersBackOffice
     config.data_retention_years = ENV["DATA_RETENTION_YEARS"] || 7
 
     # Companies House config
-    config.companies_house_host = ENV["WCRS_COMPANIES_HOUSE_URL"] || "https://api.companieshouse.gov.uk/company/"
     config.companies_house_api_key = ENV["WCRS_COMPANIES_HOUSE_API_KEY"]
+
+    config.companies_house_host =
+      if ENV["WCRS_MOCK_ENABLED"].to_s.downcase == "true"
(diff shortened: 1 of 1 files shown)


PaulDoyle-DEFRA (approved):

config/application.rb
```diff
@@ -71,9 +71,15 @@ class Application < Rails::Application
     config.data_retention_years = ENV["DATA_RETENTION_YEARS"] || 7
 
     # Companies House config
-    config.companies_house_host = ENV["WCRS_COMPANIES_HOUSE_URL"] || "https://api.companieshouse.gov.uk/company/"
     config.companies_house_api_key = ENV["WCRS_COMPANIES_HOUSE_API_KEY"]
 
+    config.companies_house_host =
+      if ENV["WCRS_MOCK_ENABLED"].to_s.downcase == "true"
+        ENV["WCRS_MOCK_FO_COMPANIES_HOUSE_URL"]
```
PaulDoyle-DEFRA, line 78: Does this need check if `ENV["WCRS_MOCK_FO_COMPANIES_HOUSE_URL"]` is defined, or is it safe to assume that for all environments where the mocks will be used?

tobyprivett: If the mock URL is defined, then  - if this is blank - it will break. But it's the pattern already used here https://github.com/DEFRA/waste-carriers-back-office/blob/a5ec1bcb90181ecfb865e8f7af5de04ea7dab552/config/initializers/waste_carriers_engine.rb#L12

We are just enabling an additional config variables

PR #1374, opened 2022-07-06
[1995] guard against empty paymentMethod

d-a-v-e:
https://eaflood.atlassian.net/browse/RUBY-1995


app/presenters/base_registration_presenter.rb
```diff
@@ -79,6 +79,12 @@ def latest_order
     finance_details&.orders&.order_by(dateCreated: :desc)&.first
   end
 
+  def latest_order_payment_method
```
PaulDoyle-DEFRA, line 82: This seems like a very specific method to be adding to the base presenter. If you're expecting it to be used in views other than `finance_details` maybe call it `order_payment_method`, acting on any `order` and not just the latest one; else move it to a specific presenter?

PR #1375, opened 2022-07-06
[1895]  Provide translation for payment_type

d-a-v-e:
https://eaflood.atlassian.net/browse/RUBY-1895

### DIFF
--- a/config/locales/shared/payment_types.en.yml
+++ b/config/locales/shared/payment_types.en.yml
@@ -11,4 +11,4 @@ en:
       WRITEOFFSMALL: "Write off (small)"
       WRITEOFFLARGE: "Write off (Form L)"
       REVERSAL: "Reversal"
-      GOVPAY: "Card (Govpay)"
+      GOVPAY: "Card"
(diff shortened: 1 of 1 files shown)


config/locales/shared/payment_types.en.yml
```diff
@@ -11,3 +11,4 @@ en:
       WRITEOFFSMALL: "Write off (small)"
       WRITEOFFLARGE: "Write off (Form L)"
       REVERSAL: "Reversal"
+      GOVPAY: "Card (Govpay)"
```
PaulDoyle-DEFRA, line 14: @d-a-v-e this needs to be checked with the business users - might need to say "Card" just like Worldpay. Let's discuss at standup.

PaulDoyle-DEFRA: I still think this needs to be confirmed, it might need to be just "Card". @timstone your thoughts?

timstone: One for Rob Bond I think, I can see an argument for having just card and one for having a reference to govpay (as it differentiates between current Worldpay payments)

PR #1399, opened 2022-07-29
Feature/nil ad email

PaulDoyle-DEFRA:
This change adds a service and Rake task to clear all NCCC contact_email addresses, including a range of known typos.
It also modifies the criteria for sending a letter vs an email to be based on the contact_email being nil, as opposed to matching a specific NCCC email address.
In addition, it removes configuration and exceptions related to the specific assisted digital email address.
https://eaflood.atlassian.net/browse/RUBY-1998


app/views/ad_reminder_letters/bulk.html.erb
```diff
@@ -1,27 +0,0 @@
-<!doctype html>
```
Beckyrose200, line 1: What was this file originally doing? It looks like a template for the reminder letters but was that not done through notify?

config/locales/resend_renewal_email.en.yml
```diff
@@ -2,9 +2,6 @@ en:
   resend_renewal_email:
     messages:
       success: Renewal email sent to %{email}
-      assisted:
-        heading: Sorry, there has been a problem re-sending the renewal email.
-        details: You have requested to resend a renewal email to the contact email address, but this address matches %{email}.
       error:
         heading: We could not send an email to %{email}
```
Beckyrose200, line 6: Will this catch errors if the email is nil? So will it say "We could not send an email to (nil)"?

spec/services/clear_nccc_contact_emails_service_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+      context "lower case (#{base_contact_email})" do
+        let(:contact_email) { base_contact_email.downcase }
+
+        it "sets contact_email to nil" do
+          expect { subject }.to change { registration.reload.contact_email }.to(nil)
+        end
+      end
+
+      # context "mixed case (#{base_contact_email})" do
```
Beckyrose200, line 84: Is this meant to be hashed out?

.github/workflows/ci.yml
```diff
@@ -26,9 +26,6 @@ jobs:
       AWS_WEEKLY_EXPORT_ACCESS_KEY_ID: 5432
       AWS_WEEKLY_EXPORT_SECRET_ACCESS_KEY: 5432
       AWS_WEEKLY_EXPORT_BUCKET: 5432
-      AWS_LETTERS_EXPORT_ACCESS_KEY_ID: 9012
```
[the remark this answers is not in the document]
PaulDoyle-DEFRA: Only one is being removed. It was part of the bulk letter generation process that was used before Notify was adopted, and was overlooked during the cleanup at that time.

app/views/ad_reminder_letters/bulk.html.erb
PaulDoyle-DEFRA: Yes, that's done through Notify now. It was previously done within the Rails application, and this was also missed during the cleanup when Notify was adopted.

config/locales/resend_renewal_email.en.yml
Beckyrose200: I read the rest of the file and can see this isn't the case so ignore me!

PaulDoyle-DEFRA: If email is nil, a `MissingContactEmailError` will be raised and `ResendRenewalEmailController` will catch it and display an error message.

spec/services/clear_nccc_contact_emails_service_spec.rb
PaulDoyle-DEFRA: No! Uncommenting....

Beckyrose200 (approved):

PR #1406, opened 2022-08-08
Adding in success and failure banner for refresh button

Beckyrose200:


app/controllers/refresh_companies_house_name_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+    redirect_back(fallback_location: registration_path(reg_identifier))
+  end
+
+  private
+
+  def authenticate_user!
+    authorize! :refresh_company_name, WasteCarriersEngine::Registration
+  end
+
+  def success_message
```
PaulDoyle-DEFRA, line 27: I think methods for these one-liners might be overkill - they are each called only once and are quite short.

PaulDoyle-DEFRA (approved):

config/locales/refresh_companies_house_name.en.yml
```diff
@@ -0,0 +1,6 @@
+en:
+  refresh_companies_house_name:
+    messages:
+      success: You have successfully refreshed the Companies House information.
+      failure: There was an error refreshing the Companies House information. Try again later
```
PaulDoyle-DEFRA, line 5: Add a full stop at the end.

PaulDoyle-DEFRA (approved):

config/locales/refresh_companies_house_name.en.yml
Beckyrose200: The wireframe doesn't include one, still want me to add it?

PaulDoyle-DEFRA: Best check with Will.

PaulDoyle-DEFRA (approved):

PR #1410, opened 2022-08-15
data agent role

PaulDoyle-DEFRA:
This creates a `data_agent` back-office user role with read-only access.
https://eaflood.atlassian.net/browse/RUBY-2052


Beckyrose200 (approved):
👍

PR #1417, opened 2022-08-23
Feature/cbd user

PaulDoyle-DEFRA:
This change renames the `import_convictions_data` back-office user role to `cbd_user`. It includes a rake task to rename the current role for any existing `import_convictions_data` users.
https://eaflood.atlassian.net/browse/RUBY-2029


kennyevil (approved):
All looks good. I was thinking that it might be worth adding a task to reverse the role rename if we need to roll back the changes for whatever reason, but that would be easy enough to create if it's needed.

PR #1464, opened 2022-10-24
Allow cbd_user to invite data_agents

PaulDoyle-DEFRA:
This change allows cbd_users to invite other users with the data_agent role (only).
https://eaflood.atlassian.net/browse/RUBY-2144


PaulDoyle-DEFRA:
This branch should not be merged until after Govpay has been deployed.

PR #1472, opened 2022-10-27
EPR pending convictions check

PaulDoyle-DEFRA:
This change includes renewing registrations which are pending convictions checks in the EPR export.
https://eaflood.atlassian.net/browse/RUBY-2113

### DIFF
--- a/spec/models/reports/epr_serializer_spec.rb
+++ b/spec/models/reports/epr_serializer_spec.rb
@@ -18,5 +18,55 @@ module Reports
         expect(result).not_to include(expired.reg_identifier)
       end
     end
+
+    describe "#scope" do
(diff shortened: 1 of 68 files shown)


spec/models/reports/epr_serializer_spec.rb
```diff
@@ -18,5 +18,55 @@ module Reports
         expect(result).not_to include(expired.reg_identifier)
       end
     end
+
+    describe "#scope" do
+      context "when a registration is pending a conviction check" do
```
timstone, line 23: Should this be, `when a renewal is pending a conviction check `?

timstone (approved):
Just one query on a spec

PR #1483, opened 2022-11-18
Show correct user status in user export CSV

endofunky:
The status precedence has been incorrect in #1482. With this commit we now correctly shows the active/deactivated status as well as uses the presence of an invitation token to determine if an active user's invitation is still pending, being now in line with what's shown on the users index page.

### DIFF
--- a/spec/models/reports/user_export_serializer_spec.rb
+++ b/spec/models/reports/user_export_serializer_spec.rb
@@ -15,8 +15,8 @@ RSpec.describe Reports::UserExportSerializer do
   let(:invited_user) do
     create(
       :user,
-      active: false,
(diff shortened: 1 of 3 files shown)


spec/models/reports/user_export_serializer_spec.rb
```diff
@@ -15,8 +15,8 @@
   let(:invited_user) do
     create(
       :user,
-      active: false,
-      invitation_sent_at: Time.at(0).utc,
+      active: true,
```
PaulDoyle-DEFRA, line 18: Is the user active as soon as invited?

PaulDoyle-DEFRA (approved):

PR #1584, opened 2023-04-12
Restore registrations - bug fixes

PaulDoyle-DEFRA:
https://eaflood.atlassian.net/browse/RUBY-2410

### DIFF
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -148,7 +148,7 @@ module ActionLinksHelper
   def display_restore_registration_link_for?(resource)
     return false unless a_registration?(resource)
     return false unless can?(:restore, WasteCarriersEngine::Registration)
-    return false if resource.expired?
+    return false if resource.expires_on.present? && WasteCarriersEngine::ExpiryCheckService.new(resource).expired?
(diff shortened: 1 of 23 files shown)


app/helpers/action_links_helper.rb
```diff
@@ -148,7 +148,7 @@ def display_transfer_link_for?(resource)
   def display_restore_registration_link_for?(resource)
     return false unless a_registration?(resource)
     return false unless can?(:restore, WasteCarriersEngine::Registration)
-    return false if resource.expired?
+    return false if resource.expires_on.present? && WasteCarriersEngine::ExpiryCheckService.new(resource).expired?
```
jjromeo, line 151: It seems strange that the presence check isn't inside the WasteCarriersEngine::ExpiryCheckService

jjromeo (approved):

app/helpers/action_links_helper.rb
PaulDoyle-DEFRA: Yes, the ExpiryCheckService assumes it is present. Not sure about the consequences elsewhere of changing that assumption within the service, so I added the check here.

PR #1593, opened 2023-04-26
Run area lookup task every day

jjromeo:


lib/tasks/lookups.rake
```diff
@@ -0,0 +1,36 @@
...
+
+      WasteCarriersEngine::Registration.where("address.area": nil).each do |registration|
+        break if counter >= address_limit
+
+        missing_area_addresses = registration.addresses.missing_area.with_postcode
+        missing_area_count = missing_area_addresses.count
+        counter += missing_area_count
+
+        # Limit the number of addresses added if the counter goes over the limit
+        missing_area_count -= (counter - address_limit) if counter > address_limit
```
PaulDoyle-DEFRA, line 22: Why not just set it to `address_limit`?

PR #1607, opened 2023-05-05
Display area in back office

jjromeo:


app/presenters/base_registration_presenter.rb
```diff
@@ -103,6 +104,10 @@ def display_last_modifed
     metaData.last_modified.to_datetime
   end
 
+  def ea_area
+    registered_address.area || "Pending"
```
PaulDoyle-DEFRA, line 108: Minor: Text should probably be defined in config/locales

PaulDoyle-DEFRA (approved):

PR #1616, opened 2023-05-19
Feature/pending refunds

PaulDoyle-DEFRA:
This adds support for pending govpay refunds and allows a back-office user to request an update of refund status.
https://eaflood.atlassian.net/browse/RUBY-2433

### DIFF
--- a/app/services/govpay_refund_details_service.rb
+++ b/app/services/govpay_refund_details_service.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+class GovpayRefundDetailsService < WasteCarriersEngine::BaseService
+  include WasteCarriersEngine::CanSendGovpayRequest
+
+  def run(refund_id:)
+    @refund = GovpayFindPaymentService.run(payment_id: refund_id)
+    raise ArgumentError, "Invalid refund id #{refund_id}" if refund.nil?
+
+    payment_id = refund.refunded_payment_govpay_id
+    # we don't use the payment value but this will force a suitable exception if not found
+    @payment = GovpayFindPaymentService.run(payment_id:)
+    raise ArgumentError, "Invalid refunded payment id #{payment_id}" if payment.nil?
+
+    response
+  rescue StandardError => e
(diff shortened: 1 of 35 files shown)


brujeo (approved):

app/services/govpay_refund_details_service.rb
```diff
@@ -0,0 +1,39 @@
...
+  rescue StandardError => e
+    Rails.logger.error "#{e.class} error in Govpay refund details service for payment  " \
+                       "#{payment_id}, refund #{refund_id}"
+    Airbrake.notify(e, message: "Error in Govpay refund details service", payment_id:, refund_id:)
+    raise e
+  end
+
+  private
+
+  attr_reader :payment, :refund
```
jjromeo, line 25: Do we really need the attr reader for payment since it isn't really used?

app/services/govpay_refund_details_service.rb
```diff
@@ -0,0 +1,39 @@
...
+class GovpayRefundDetailsService < WasteCarriersEngine::BaseService
+  include WasteCarriersEngine::CanSendGovpayRequest
+
+  def run(refund_id:)
+    @refund = GovpayFindPaymentService.run(payment_id: refund_id)
+    raise ArgumentError, "Invalid refund id #{refund_id}" if refund.nil?
+
+    payment_id = refund.refunded_payment_govpay_id
+    # we don't use the payment value but this will force a suitable exception if not found
+    @payment = GovpayFindPaymentService.run(payment_id:)
```
jjromeo, line 12: Is this argument right? Not sure I understand this syntax, what exactly is the argument?

brujeo (approved):

app/services/govpay_refund_details_service.rb
PaulDoyle-DEFRA: It's used on line 29

app/services/govpay_refund_details_service.rb
PaulDoyle-DEFRA: It's the equivalent of `(payment_id: payment_id)`. Rubocop prefers the abbreviated syntax.

PR #1626, opened 2023-05-25
govpay refund details - moto payments

PaulDoyle-DEFRA:
This change fixes an issue with the govpay refund details service for MOTO payments. It also prevents displaying the Refund button on the finance-details page if there is a pending govpay payment.
https://eaflood.atlassian.net/browse/RUBY-2433

### DIFF
--- a/app/helpers/action_links_helper.rb
+++ b/app/helpers/action_links_helper.rb
@@ -68,6 +68,9 @@ module ActionLinksHelper
     # Do not display the link if there is a pending refund
     return false unless resource.payments.find(&:refund?).nil?
 
+    # Do not display the link if there is a pending refund
+    return false unless resource.payments.find { |payment| payment.payment_type == "REFUND" }.nil?
+
     can?(:refund, resource)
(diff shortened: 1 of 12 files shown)


app/helpers/action_links_helper.rb
```diff
@@ -65,6 +65,9 @@ def display_refund_link_for?(resource)
     # Display the link only if there is at least one payment of the relevant type
     return false if resource.payments.find { |payment| payment.payment_type == "GOVPAY" }.nil?
```
brujeo, line 66: I think it would be safer not to hardcode payment types and use convenience methods like payment.govpay? or payment.refund?

`return false if resource.payments.find { |payment| payment.payment_type == "GOVPAY" }.nil?`
=>
`return false if resource.payments.find { |payment| payment.govpay? }.nil?`

brujeo (approved):

brujeo (approved):

brujeo (approved):

PR #1651, opened 2023-06-16
Fix/area query

PaulDoyle-DEFRA:
This change adds a feature toggle for the missing_area rake task. It also resolves some intermittent unit test failures.
https://eaflood.atlassian.net/browse/RUBY-2558


app/presenters/notify_renewal_presenter.rb
jjromeo, line 1: This file should be deleted, will be added back in https://github.com/DEFRA/waste-carriers-back-office/pull/1647

jjromeo (approved):

PR #1726, opened 2023-10-10
[RUBY-2667] transfer link should be disabled once frontend users are removed

brujeo:
https://eaflood.atlassian.net/browse/RUBY-2667


app/helpers/action_links_helper.rb
```diff
@@ -153,11 +153,8 @@ def display_renew_link_for?(resource)
     resource.can_start_renewal?
   end
 
-  def display_transfer_link_for?(resource)
-    return false if WasteCarriersEngine::FeatureToggle.active?(:block_front_end_logins)
-    return false unless display_registration_links?(resource)
-
-    can?(:transfer_registration, WasteCarriersEngine::Registration)
+  def display_transfer_link_for?(_resource)
```
PaulDoyle-DEFRA, line 156: I think this should just be removed

brujeo: removed

PaulDoyle-DEFRA: It's still there

brujeo: looks like my previous commit hasn't been pushed. That code is gone now

PaulDoyle-DEFRA (approved):

PR #1732, opened 2023-10-17
[RUBY-2730] Improve specs so that they don't fail depending on environment config

jjromeo:
Also raise log level from debug to log to avoid lengthy govpay logs in
specs

### DIFF
--- a/spec/services/govpay_refund_details_service_spec.rb
+++ b/spec/services/govpay_refund_details_service_spec.rb
@@ -7,7 +7,6 @@ module WasteCarriersEngine
 
     describe "#run" do
 
-      let(:govpay_host) { "https://publicapi.payments.service.gov.uk" }
       let(:registration) { create(:registration, finance_details: build(:finance_details, :has_overpaid_order_and_payment_govpay)) }
       let(:original_payment) { registration.finance_details.payments.first }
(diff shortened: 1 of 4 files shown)


spec/services/govpay_refund_details_service_spec.rb
```diff
@@ -7,17 +7,13 @@ module WasteCarriersEngine
...
 
-      let(:govpay_host) { "https://publicapi.payments.service.gov.uk" }
       let(:registration) { create(:registration, finance_details: build(:finance_details, :has_overpaid_order_and_payment_govpay)) }
       let(:original_payment) { registration.finance_details.payments.first }
       let(:govpay_payment_id) { original_payment.govpay_id }
       let(:refund) { build(:payment, :govpay_refund_pending, refunded_payment_govpay_id: original_payment.govpay_id) }
       let(:govpay_refund_id) { refund.govpay_id }
 
       before do
-        allow(WasteCarriersEngine.configuration).to receive(:host_is_back_office?).and_return(true)
```
PaulDoyle-DEFRA, line 18: isn't this needed so we can test that the service runs only in the back office?

PaulDoyle-DEFRA (approved):

PR #1767, opened 2023-11-10
Bump govpay gem version

PaulDoyle-DEFRA:
Adjust govpay config in unit tests and bump defra-ruby-govpay gem version.
https://eaflood.atlassian.net/browse/RUBY-2793

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -61,8 +61,6 @@ gem "waste_carriers_engine",
 # manage feature toggle (create / update / delete) from the back-office.
 gem "defra_ruby_features"
 
-gem "defra_ruby_govpay"
-
 # Use the defra ruby mocks engine to add support for mocking external services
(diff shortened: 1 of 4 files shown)


Gemfile
```diff
@@ -61,8 +61,6 @@ gem "waste_carriers_engine",
 # manage feature toggle (create / update / delete) from the back-office.
 gem "defra_ruby_features"
 
-gem "defra_ruby_govpay"
-
```
brujeo, line 65: Is defra_ruby_govpay required by waste-carriers-engine now?

brujeo (dismissed):

Gemfile
PaulDoyle-DEFRA: Yes - it always was in the gemspec, so not needed here

brujeo (approved):

PR #1774, opened 2023-11-16
Fix/ruby 2793 govpay back office

PaulDoyle-DEFRA:
Align with the updated DefraRubyGovpay interface
https://eaflood.atlassian.net/browse/RUBY-2793

### DIFF
--- a/app/services/govpay_refund_details_service.rb
+++ b/app/services/govpay_refund_details_service.rb
@@ -22,9 +22,13 @@ class GovpayRefundDetailsService < WasteCarriersEngine::BaseService
 
   attr_reader :payment, :refund
 
+  def defra_ruby_govpay_api
+    DefraRubyGovpay::API.new(host_is_back_office: true)
+  end
+
   def request
     @request ||=
-      DefraRubyGovpayAPI.send_request(
+      defra_ruby_govpay_api.send_request(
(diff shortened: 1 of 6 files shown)


app/services/govpay_refund_details_service.rb
```diff
@@ -22,9 +22,13 @@ def run(refund_id:)
 
   attr_reader :payment, :refund
 
+  def defra_ruby_govpay_api
+    DefraRubyGovpay::API.new(host_is_back_office: true)
```
brujeo, line 26: I don't think we want new govpay API connection to be created on each request. Probably worth storing and reusing existing connection (singleton)

app/services/govpay_refund_service.rb
```diff
@@ -50,9 +50,13 @@ def params
     }
   end
 
+  def defra_ruby_govpay_api
+    DefraRubyGovpay::API.new(host_is_back_office: true)
```
brujeo, line 54: same here, it would be good to reuse the connection

brujeo (approved):

app/services/govpay_refund_details_service.rb
PaulDoyle-DEFRA: As discussed - the govpay gem creates a new request each time so there would be little or no advantage to reusing the API object.

PR #1782, opened 2023-11-23
RUBY 2765 wcr bo call recording to be paused before card payment

jjromeo:
- [RUBY-2765] Add `defra_ruby_storm` gem and update `Gemfile.lock` with its dependencies.
- [RUBY-2765] Add call recording control and integrate with payment forms
- [RUBY-2765] Rubocop fixes:
- [RUBY-2765] Add specs for CallRecordingService pause method with various scenarios

[RUBY-2765]: https://eaflood.atlassian.net/browse/RUBY-2765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RUBY-2765]: https://eaflood.atlassian.net/browse/RUBY-2765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RUBY-2765]: https://eaflood.atlassian.net/browse/RUBY-2765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RUBY-2765]: https://eaflood.atlassian.net/browse/RUBY-2765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.rubocop.yml
```diff
@@ -29,3 +29,6 @@ RSpec/DescribeClass:
 # Using 'let(:registration1)' in a unit test isn't necessarily unreasonable.
 RSpec/IndexedLet:
   Enabled: false
+
+Rails/LexicallyScopedActionFilter:
```
PaulDoyle-DEFRA, line 33: Does this need to be disabled for all files? If it's just for a few files, can the before_action be refactored to comply with the cop?

Gemfile
```diff
@@ -57,6 +59,8 @@ gem "waste_carriers_engine",
     git: "https://github.com/DEFRA/waste-carriers-engine",
     branch: "main"
 
+# Get waste carriers engine from local path for development
```
PaulDoyle-DEFRA, line 62: Stray comment

app/controllers/user_invitations_controller.rb
```diff
@@ -3,7 +3,6 @@
 class UserInvitationsController < Devise::InvitationsController
   include UsersHelper
 
-  # rubocop:disable Rails/LexicallyScopedActionFilter
   before_action :authorize, only: %i[new create]
   # rubocop:enable Rails/LexicallyScopedActionFilter
```
PaulDoyle-DEFRA, line 7: Should be deleted along with the rubocop:disable line

config/storm_email_to_username_mappings.yml
```diff
@@ -0,0 +1,2 @@
+agency-user@wcr.gov.uk: CarrierTest
```
PaulDoyle-DEFRA, line 1: Isn't this for temporary local use only? In which case do not commit

jjromeo: This is also for use in the development environment

.rubocop.yml
jjromeo: Unfortunately in trying to fix this I am now facing a conflict between two RuboCop rules:

Rails/LexicallyScopedActionFilter: This rule checks that actions are explicitly defined in the controller where a filter is applied. In my case, the new action is not explicitly defined in CopyCardsPaymentFormsController and the other two controllers, but I have a before_action filter that applies to it. This is why RuboCop is raising a warning.

Lint/UselessMethodDefinition: This rule checks for method definitions that don't do anything but call super with the same arguments that were passed to the method. If I define the new action in CopyCardsPaymentFormsController to satisfy the Rails/LexicallyScopedActionFilter rule, and the new action only calls super, then RuboCop will raise a warning for the Lint/UselessMethodDefinition rule.

I therefore have to disable one of these cops, unless you have another idea on how to resolve?

PaulDoyle-DEFRA: Awkward. Maybe just disable for the specific method rather than disabling globally.

PaulDoyle-DEFRA (approved):

PR #1808, opened 2023-12-21
[RUBY-2795] view communication history page

brujeo:
https://eaflood.atlassian.net/browse/RUBY-2795

### DIFF
--- a/spec/requests/communication_records_spec.rb
+++ b/spec/requests/communication_records_spec.rb
@@ -12,34 +12,75 @@ RSpec.describe "Communication Records" do
 
   describe "GET /bo/registrations/:reg_identifier/communication_records/" do
     context "when communication history is present" do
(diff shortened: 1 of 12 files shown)


spec/requests/communication_records_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+        get "/bo/registrations/#{registration.reg_identifier}/communication_records"
+
+        expect(response).to render_template(:index)
+        expect(response).to have_http_status(:ok)
+        expect(response.body).to include("Communication history")
+        expect(response.body).to include("No results found")
+      end
+    end
+
+    context "when user has no access or registration does not exist" do
```
PaulDoyle-DEFRA, line 42: I think this should be two different examples with different setup as the conditions are different.

PaulDoyle-DEFRA (approved):

PR #1865, opened 2024-02-21
Ruby 2962 improve compatibility of conviction data import with current process used

jjromeo:


spec/requests/conviction_imports_spec.rb
```diff
@@ -61,6 +61,54 @@
...
+        let(:utf_8_csv_with_data) { fixture_file_upload(Rails.root.join("spec/fixtures/files/utf-8-with-data.csv"), "text/csv") }
+        let(:utf_8_csv_with_data_and_bom) { fixture_file_upload(Rails.root.join("spec/fixtures/files/valid_entities-utf8.csv"), "text/csv") }
+
+        context "with a file with only headers" do
+          context "when the file is utf-8 encoded" do
+            it "redirects to the results page and displays a flash message" do
+              post "/bo/import-convictions", params: { file: empty_utf_8_csv }
+
+              expect(response).to redirect_to(bo_path)
+              expect(flash[:success]).to match(/Convictions data has been updated successfully. \d+ records in database./)
```
PaulDoyle-DEFRA, line 76: Should the message not state that no rows were loaded? The user is unlikely to have intended to drop all existing convictions.

PR #1894, opened 2024-03-25
[RUBY-2888] Add upper tier check for call recording resumption and update tests

jjromeo:


app/controllers/concerns/can_resume_call_recording.rb
```diff
@@ -4,19 +4,24 @@ module CanResumeCallRecording
...
-                               { error: t("shared.call_recording_banner.call_resuming.error") }
-                             end
+    resume_call_recording
   end
 
   private
 
+  def should_resume_call_recording?
+    WasteCarriersEngine::FeatureToggle.active?(:control_call_recording) &&
+      (!@transient_registration.lower_tier? || @transient_registration.is_a?(OrderCopyCardsRegistration))
```
PaulDoyle-DEFRA, line 16: If it's a copy cards order the registration must be upper tier - why check both?

PaulDoyle-DEFRA (approved):

PR #1897, opened 2024-03-28
Feature/ruby 2973 unsubscribe

PaulDoyle-DEFRA:
This change prevents renewal reminder emails being sent to users who have unsubscribed.
It also adds a database index on registration unsubscribe tokens to avoid slow searches.
Finally, it includes a change to rubocop configuration and a large number number of associated code changes, primarily removing `rubocop:disable` entries.
https://eaflood.atlassian.net/browse/RUBY-2973


app/services/reports/defra_quarterly_stats_service.rb
```diff
@@ -2,10 +2,10 @@
 
 # This is a minimalist implementation to meet an urgent requirement. It is expected to be enhanced later.
 module Reports
+  # rubocop:disable Layout/LineLength
```
brujeo, line 5: Don't we need # rubocop:enable comment afterwards?

brujeo (approved):

PR #1905, opened 2024-04-09
Feature/ruby 2973 unsubscribe

PaulDoyle-DEFRA:
https://eaflood.atlassian.net/browse/RUBY-2973

### DIFF
--- a/app/services/notify/renewal_reminder_email_service.rb
+++ b/app/services/notify/renewal_reminder_email_service.rb
@@ -2,7 +2,7 @@
 
 module Notify
   class RenewalReminderEmailService < ::WasteCarriersEngine::Notify::BaseSendEmailService
-    TEMPLATE_ID = "51cfcf60-7506-4ee7-9400-92aa90cf983c"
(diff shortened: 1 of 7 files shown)


app/services/notify/renewal_reminder_email_service.rb
```diff
@@ -14,7 +14,7 @@ def template_id
     def notify_options
       {
         email_address: ContactEmailValidatorService.run(@registration),
-        template_id: "6d20d279-ba79-4fe0-9868-fb09c4ae7733",
+        template_id:,
```
brujeo, line 17: Do you know why template IDs in TEMPLATE_ID constant and notify_options were different before? What you've done now looks correct, but I'd like to double check if I missed something

brujeo (approved):

app/services/notify/renewal_reminder_email_service.rb
PaulDoyle-DEFRA: It's not clear to me why template_id and TEMPLATE_ID were set up that way. I had updated one value but not noticed the other. I think we should try to defined each id once and once only. I'll create a tech debt item to clean all this up.

PR #1925, opened 2024-04-24
fix/RUBY-3074_edit_bank_transfer_form_back

PaulDoyle-DEFRA:
Fix back button on edit_bank_transfer_form
https://eaflood.atlassian.net/browse/RUBY-3074


app/views/edit_bank_transfer_forms/new.html.erb
```diff
@@ -1,4 +1,5 @@
-<%= render("waste_carriers_engine/shared/back", token: @edit_bank_transfer_form.token) %>
+<% byebug %>
```
brujeo, line 1: This needs to go

brujeo (changes requested):

app/views/edit_bank_transfer_forms/new.html.erb
PaulDoyle-DEFRA: Ah byebug doesn't do erb....

brujeo (approved):

PR #1933, opened 2024-05-07
Chore/rails logger

PaulDoyle-DEFRA:
- Align with engine Logger changes for log rotation and filtering
- Allow log severity threshold to be configurable
- Simplify FeatureToggle "allow" handling in specs
https://eaflood.atlassian.net/browse/RUBY-2862
https://eaflood.atlassian.net/browse/RUBY-3031

### DIFF
--- a/spec/requests/copy_cards_order_completed_forms_spec.rb
+++ b/spec/requests/copy_cards_order_completed_forms_spec.rb
@@ -7,11 +7,6 @@ RSpec.describe "CopyCardsOrderCompletedForms" do
   describe "GET new_copy_cards_order_completed_form_path" do
     let(:path) { new_copy_cards_order_completed_form_path(transient_registration.token) }
 
-    before do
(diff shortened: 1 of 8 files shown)


spec/requests/copy_cards_order_completed_forms_spec.rb
```diff
@@ -7,11 +7,6 @@
   describe "GET new_copy_cards_order_completed_form_path" do
     let(:path) { new_copy_cards_order_completed_form_path(transient_registration.token) }
 
-    before do
```
brujeo, line 10: Not sure how this is related to logging refatoring?

PaulDoyle-DEFRA: When adding the FeatureToggle for log filtering I had to add lots of stubs for unrelated tests because of the way other FeatureToggle'd features were set up. So rather than add to the existing mess I simplified it.

PR #1988, opened 2024-07-15
Chore/webhook signature api

PaulDoyle-DEFRA:
This adds a helper API to support testing of Gov.UK Pay callbacks. The API generates a signature for the request payload; this signature can subsequently be sent in the header of the request to the actual webhook endpoint.
The API should be available only when the `:api` feature toggle is active.
https://eaflood.atlassian.net/browse/RUBY-3143


spec/requests/api/govpay_webhooks_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  before { allow(WasteCarriersEngine::FeatureToggle).to receive(:active?).with(:api).and_return(true) }
+
+  describe "POST /bo/api/govpay_webhooks/signature" do
+    it "returns a valid hexadecimal value" do
+      params = { foo: :bar }
+
+      post "/bo/api/govpay_webhooks/signature", params: params.to_json, headers: { "CONTENT_TYPE" => "application/json" }
+
+      expect(response.body.match(/^[0-9A-F]+$/i)).to be_present
```
brujeo, line 15: This does test only that some hexadecimal response is being returned, but does not test whether that response is correct or not.

app/controllers/api/govpay_webhooks_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
+
+    def hmac_digest(body)
+      digest = OpenSSL::Digest.new("sha256")
+      OpenSSL::HMAC.hexdigest(digest, webhook_signing_secret, body)
+    end
+
+    def skip_auth_on_this_controller?
+      true
+    end
```
brujeo, line 28: I think this needs to be moved into helper so that it can be reused while validating webhook signature. Otherwise you'll have to duplicate this code

PaulDoyle-DEFRA: Wouldn't that be circular? The spec using the same substantive code as the implementation to test the behaviour of the implementation?

spec/requests/api/govpay_webhooks_spec.rb
brujeo: It could be something as simple as this

params = {param1: 'one', param2: 'two}
post "/bo/api/govpay_webhooks/signature", params: params.to_json, headers: { "CONTENT_TYPE" => "application/json" }
expect(response.body).to eq('345g2dee3qw2e23321')

brujeo (dismissed):

PR #2032, opened 2024-10-07
Fix/ruby 3411 determine address services

PaulDoyle-EA:
This change:
- Reinstates, renames and refactors the (formerly engine) services for determining areas and easting/northing values
- Splits the missing_address_attributes task into the original missing_area task (renamed missing_ea_area for clarity, and refactored) and a new missing_easting_northing service
- Adds the latter service to the schedule
https://eaflood.atlassian.net/browse/RUBY-3411

### DIFF
--- a/app/services/geographic/determine_ea_area_service.rb
+++ b/app/services/geographic/determine_ea_area_service.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+require "defra_ruby/area"
+
+module Geographic
+  class DetermineEaAreaService < WasteCarriersEngine::BaseService
+    def run(easting:, northing:)
+      response = DefraRuby::Area::PublicFaceAreaService.run(easting, northing)
+
+      return response.areas.first.long_name if response.successful?
+      return "Outside England" if response.error.instance_of?(DefraRuby::Area::NoMatchError)
+
+      handle_error(response.error, easting, northing)
+      nil
+    end
+
+    private
+
+    def handle_error(error, easting, northing)
+      Airbrake.notify(error, easting: easting, northing: northing)
+      Rails.logger.error "Area lookup failed:\n #{error}"
+    end
+  end
+end
--- a/app/services/geographic/update_ea_area_service.rb
+++ b/app/services/geographic/update_ea_area_service.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+module Geographic
+  class UpdateEaAreaService < WasteCarriersEngine::BaseService
+    attr_reader :address, :registration, :easting, :northing
+
+    delegate :postcode, :area, to: :address
+
+    def run(registration_id:)
+      @registration = WasteCarriersEngine::Registration.find(registration_id)
+      @address = @registration.company_address
+
+      set_easting_and_northing
+      assign_area
+    end
+
+    private
+
+    def set_easting_and_northing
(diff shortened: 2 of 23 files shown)


config/schedule.rb
```diff
@@ -92,6 +92,14 @@
...
 end
 
+every :day, at: ENV["AREA_LOOKUP"] || "01:05", roles: [:db] do
+  rake_and_format "lookups:update:missing_ea_areas"
+end
+
+every :day, at: ENV["EASTING_NORTHING_LOOKUP"] || "03:30", roles: [:db] do
+  rake_and_format "lookups:update:missing_easting_northing"
+end
+
```
timstone, line 102: @PaulDoyle-EA shouldn't these be the other way round? i.e. does the area lookup need an easting and northing?

app/services/geographic/update_ea_area_service.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module Geographic
```
brujeo, line 3: I don't think this belongs to Geographic namespace. This service is not performing any geo mapping tasks, it's just calling other services and updates address area

app/services/geographic/update_easting_northing_service.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module Geographic
```
brujeo, line 3: same here, I don't think it belongs to Geographic namespace

app/services/geographic/determine_ea_area_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require "defra_ruby/area"
+
+module Geographic
+  class DetermineEaAreaService < WasteCarriersEngine::BaseService
```
brujeo, line 6: The problem I find with DetermineEaAreaService name is that it doesn't suggest what are we using as an input. And that input might be different: postcode, eating and nothing, national grid reference, etc

How about renaming these services e.g. Geographic::DetermineEaAreaFromEastingAndNorthingService,

or 

Geographic::MapEastingAndNorthingToEaAreaService
Geographic::MapPostcodeToEastingAndNorthingService

brujeo (dismissed):

config/schedule.rb
PaulDoyle-EA: Logically yes, though the area lookup retrieves easting and northing in the same way as the missing_easting_northing job so it won't make any difference in practice. I'll switch them around for the sake of tidiness.

app/services/geographic/update_ea_area_service.rb
PaulDoyle-EA: Would "addresses" be better?

app/services/geographic/determine_ea_area_service.rb
PaulDoyle-EA: The "Map..." option works for me.

app/services/geographic/update_ea_area_service.rb
brujeo: Yes. It would make sense to reserve Geographic namespace for services that do geo mapping but have no knowledge of business logic of the app. Later on, we could take Geographic into a separate gem and use it across different apps

brujeo (approved):

PR #2353, opened 2026-07-24
[RUBY-4350] Use local geospatial EA area lookup

jjromeo:
https://eaflood.atlassian.net/browse/RUBY-4350

Requires DEFRA/waste-carriers-engine#1735. Swaps `Geographic::MapEastingAndNorthingToEaAreaService` from the external DefraRuby::Area API call (gem now removed) to the engine's new `DetermineEaAreaService`, so areas come from a geospatial query against the actual boundaries. Both callers (the "Refresh EA area" action and the nightly lookups job) go through this service, so nothing else changes.

Adds a `load_ea_areas` rake task and load service, following the equivalent data load in WEX and FRAE, to import the EA area boundaries GeoJSON fixture into MongoDB, converting easting / northing to WGS84 on the way in.

Will require running `load_ea_areas` once per environment on deploy. Until then the collection is empty and lookups return "Outside England".

Gemfile is pinned to the engine feature branch for CI; to be re-pointed at main once the engine PR is merged.


app/services/geographic/map_easting_and_northing_to_ea_area_service.rb
```diff
@@ -1,18 +1,13 @@
...
     def run(easting:, northing:)
-      return if easting.blank? || northing.blank?
-
-      response = DefraRuby::Area::PublicFaceAreaService.run(easting, northing)
-
-      return response.areas.first.long_name if response.successful?
-      return "Outside England" if response.error.instance_of?(DefraRuby::Area::NoMatchError)
-
-      handle_error(response.error, easting, northing)
+      WasteCarriersEngine::DetermineEaAreaService.run(easting: easting, northing: northing)
```
brujeo, line 12: Until the collection is full, the service will mark any area as "Outside England". In order to avoid that, I'd suggest to add a safety check before the service is triggered
```
return nil if WasteCarriersEngine::EaPublicFaceArea.none?

WasteCarriersEngine::DetermineEaAreaService.run(easting: easting, northing: northing)
```

brujeo (dismissed):

PaulDoyle-EA (approved):

